PR #5, opened 2017-05-26
[WIP] Fix apt-get installation stuff

staticfloat:
Work in progress, simultaneously testing this and the new sudo-less bindeps branch

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,6 @@ deps/deps.jl
 deps/downloads
 deps/usr
 deps/src
+deps/builds
 docs/build
 docs/site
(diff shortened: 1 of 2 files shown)


staticfloat:
Alright this works on Ubuntu.  I'ma merge it.

ararslan:
Thanks! I'm surprised there isn't a single FFTW package available from APT that provides both single- and double-precision.

staticfloat:
There is, but this is the most minimal way to get what we want.  :)

ararslan:
Sounds great. Thanks again!

PR #6, opened 2017-06-03
use `@__DIR__`

musm:
Might as well since this is 0.6 only

### DIFF
--- a/src/FFTW.jl
+++ b/src/FFTW.jl
@@ -2,7 +2,7 @@ __precompile__()
 
 module FFTW
 
-const depsfile = joinpath(dirname(@__FILE__), "..", "deps", "deps.jl")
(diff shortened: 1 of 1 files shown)


src/FFTW.jl
```diff
@@ -2,7 +2,7 @@ __precompile__()
 
 module FFTW
 
-const depsfile = joinpath(dirname(@__FILE__), "..", "deps", "deps.jl")
+const depsfile = joinpath(dirname(@__DIR__), "deps", "deps.jl")
```
stevengj: I don't really understand why we do this rather than `joinpath("..", "deps", "deps.jl")`, since `include` is always relative to the path of the current file.

musm: good point I opened a pr to change it to your suggestion

PR #7, opened 2017-06-05
Use relative path for the deps include

musm:
https://github.com/JuliaMath/FFTW.jl/pull/6#discussion_r120142738

### DIFF
--- a/src/FFTW.jl
+++ b/src/FFTW.jl
@@ -2,7 +2,7 @@ __precompile__()
 
 module FFTW
 
-const depsfile = joinpath(dirname(@__DIR__), "deps", "deps.jl")
+const depsfile = joinpath("..", "deps", "deps.jl")
 if isfile(depsfile)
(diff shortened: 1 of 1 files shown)


ararslan:
I prefer using the full path but I guess this is fine

musm:
Actually I think we need to revert this.
I was debugging and running issues with ImageMagick and apparently the problem was because I reverted to using relative paths for the deps include file ! was pretty confused and this caused me a bit of grief.

musm:
cc @stevengj

stevengj:
I'm confused by this.  `include` is always relative to the current path, so what possible problems could a relative path cause?

stevengj:
(Functions *other* than `include`, like `open` etcetera, are *not* relative to the current file's path and *do* need to use absolute paths to open files in the package directory.  Maybe that's what you were seeing?)

PR #8, opened 2017-06-05
Revert "Use relative path for the deps include"

musm:
Reverts JuliaMath/FFTW.jl#7

### DIFF
--- a/src/FFTW.jl
+++ b/src/FFTW.jl
@@ -2,7 +2,7 @@ __precompile__()
 
 module FFTW
 
-const depsfile = joinpath("..", "deps", "deps.jl")
+const depsfile = joinpath(dirname(@__DIR__), "deps", "deps.jl")
 if isfile(depsfile)
     include(depsfile)
 else


stevengj:
I'm opposed to this unless we can clearly identify what problem this solves.

musm:
For reference both myself and @Allardvm observed this problem when we did not include the absolute path. We are not exactly sure but the only thing that this file contained was  the library path and dictionary entries.  FWIW this was an issue we observed in windows 

perhaps @tkelman  has seen this in the past

stevengj:
Library paths need to be absolute. Not include paths

stevengj:
(Merging workarounds without diagnosing the problem is a good way to hide bugs/misunderstandings.)

tkelman:
I think it's the `isfile` that can change behavior here, but I forget in what way

ararslan:
We could just use full paths everywhere, since they're guaranteed to work as expected.

stevengj:
Right, isfile is the problem here.

musm:
should a bug report be open to try and fix `ifile` or is that not possible?

musm:
FWIW  I believe the issue we were seeing was related to something else (i.e. I can't seem to trigger the problem with using relative paths with `include`)

I agree using relative paths should just work without having to special case cases like this which I guess can lead to hard to trace bugs

stevengj:
@musm, no, `isfile`, along with all other file operations, is supposed to be relative to the current working directory (`pwd()`), which is *not* necessarily the same as the directory of the current file.

*Only* `include` is supposed to be relative to the directory of the current *file*.

PR #9, opened 2017-06-06
Add a dependency on AbstractFFTs

### LINKED ISSUE #2: split package
The point of the dft.jl stuff is that you can plug in non-FFTW implementations (e.g. a pure-Julia FFT, FFTPACK, whatever) and it will still work with the same interface.   So, it really needs to be split into a separate package.

My suggestion would be that there be a separate package (maybe "AbstractFFTs"?) that contains dft.jl.  Then any package that provides FFT functionality would use this package.

Similarly the dsp stuff should probably be in a separate package.

ararslan:
Fixes #2


test/runtests.jl
```diff
@@ -2,6 +2,7 @@
 using FFTW
 using Base.Test
 
+importall AbstractFFTs
 importall FFTW
```
stevengj: Why do the tests need `importall` and not just `using`?

ararslan: To make sure these names get picked up rather than the ones from Base

ararslan:
Wooooooooo tests are passing now! Want to take another look, @stevengj?

stevengj:
Lgtm

ararslan:
The macOS errors don't seem related to the package, but they are strange. I wonder if there's something I need to do differently with the Travis setup for Homebrew to work?

PR #10, opened 2017-06-07
Move thread initialization to module __init__

ararslan:
This was a to-do for this day, when FFTW is in its own package

### DIFF
--- a/src/FFTW.jl
+++ b/src/FFTW.jl
@@ -47,16 +47,12 @@ end
 #      If this isn't called before the FFTW planner is created, then
 #      FFTW's threads algorithms won't be registered or used at all.
 #      (Previously, we called fftw_cleanup, but this invalidated existing
-#       plans, causing issue #19892.)
-const threads_initialized = Ref(false)
+#       plans, causing Base Julia issue #19892.)
 function __init__()
-    if !threads_initialized[]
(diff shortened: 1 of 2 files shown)


src/FFTW.jl
```diff
@@ -42,6 +42,24 @@ else
     const libfftwf_name = "libfftw3f_threads"
 end
 
+# Threads must be initialized before any FFTW planning routine.
+#   -- This initializes FFTW's threads support (defaulting to 1 thread).
+#      If this isn't called before the FFTW planner is created, then
+#      FFTW's threads algorithms won't be registered or used at all.
+#      (Previously, we called fftw_cleanup, but this invalidated existing
+#       plans, causing issue #19892.)
```
giordano: The comment refer to a ticket in Julia's repository.  Maybe update it for future reference?

src/FFTW.jl
```diff
@@ -42,6 +42,24 @@ else
...
+const threads_initialized = Ref(false)
+function __init__()
+    if !threads_initialized[]
```
stevengj: since `__init__` is always called exactly once, presumably we can get rid of the `threads_initialized` check?

ararslan:
Looking better?

PR #11, opened 2017-06-07
Remove DSP functions (to be rehomed)

ararslan:
Originally I had a different vision for how to get these things removed from Base. Since we're doing as separate packages (e.g. AbstractFFTs and FFTW), I figured it would make more sense to just go ahead and move this stuff to DSP rather than keep it here.


ararslan:
Once this and #10 are merged, I'll register the package, then set up DSP to have a dependency on this.

ararslan:
Oh whoops, we still need some of the stuff in test/dsp.jl. Don't merge yet.

ararslan:
@stevengj Any idea why Coveralls is showing 0% coverage for src/dct.jl, even with the DCT tests included?

ararslan:
Never mind, I know why. It's calling the Base functions.

PR #12, opened 2017-06-17
Update imports for FFTW removal from Base

ararslan:
At first I was going to use this as a dummy PR to test the nightly binary which now does not include FFTW, but then I realized we should probably do this anyway.

### DIFF
--- a/src/FFTW.jl
+++ b/src/FFTW.jl
@@ -13,8 +13,12 @@ import AbstractFFTs: Plan, ScaledPlan,
                      rfft_output_size, brfft_output_size,
                      plan_inv, normalization
 
-if !isdefined(Base, :FFTW)
-    export dct, idct, dct!, idct!, plan_dct, plan_idct, plan_dct!, plan_idct!
+if VERSION >= v"0.7.0-DEV.602"
+    if isdefined(Base, :FFTW)
+        import Base.FFTW: dct, idct, dct!, idct!, plan_dct, plan_idct, plan_dct!, plan_idct!
+    else
+        export dct, idct, dct!, idct!, plan_dct, plan_idct, plan_dct!, plan_idct!
+    end
 end
 
 const depsfile = joinpath(dirname(@__DIR__), "deps", "deps.jl")


ararslan:
FYI @tkelman this is where I'm experimenting with trying to fix FFTW on master. No luck yet. Seems that BinDeps isn't correctly generating the deps.jl file; when I run it locally, the generated file has nothing in the "load dependencies" section.

tkelman:
Thanks for the heads up. Does BinDeps.debug say anything interesting? How about trying to dlopen the built library?

ararslan:
I'm not familiar with `BinDeps.debug`. On my system it's just getting the binary from Homebrew, which works fine using `dlopen` directly.

tkelman:
I think there's an equivalent of this for travis to ssh into a build worker for debug purposes, but I'm having trouble finding it. The appveyor version is https://www.appveyor.com/docs/how-to/rdp-to-build-worker/

ararslan:
I just commented out the Homebrew part locally to force it to build. The files seem to be where they need to be, and `dlopen` works fine on the built library, but the deps.jl file still isn't coming out right: it's missing the call to `@checked_lib`:
```julia
# This is an auto-generated file; do not edit

# Pre-hooks

# Macro to load a library
macro checked_lib(libname, path)
    ((VERSION >= v"0.4.0-dev+3844" ? Base.Libdl.dlopen_e : Base.dlopen_e)(path) == C_NULL) && error("Unable to load \n\n$libname ($path)\n\nPlease re-run Pkg.build(package), and restart Julia.")
    quote const $(esc(libname)) = $path end
end

# Load dependencies

# Load-hooks

```

tkelman:
That's usually an indication that dlopen failed?

ararslan:
Looks like it generated the expected structure and files though:
```
.
└── deps
    ├── build.jl
    ├── builds
    ├── deps.jl
    ├── downloads
    ├── src
    └── usr

deps/usr
├── bin
│   ├── fftw-wisdom
│   ├── fftw-wisdom-to-conf
│   └── fftwf-wisdom
└── lib
    ├── libfftw3.3.dylib
    ├── libfftw3.a
    ├── libfftw3.dylib -> libfftw3.3.dylib
    ├── libfftw3.la
    ├── libfftw3_threads.3.dylib
    ├── libfftw3_threads.a
    ├── libfftw3_threads.dylib -> libfftw3_threads.3.dylib
    ├── libfftw3_threads.la
    ├── libfftw3f.3.dylib
    ├── libfftw3f.a
    ├── libfftw3f.dylib -> libfftw3f.3.dylib
    ├── libfftw3f.la
    ├── libfftw3f_threads.3.dylib
    ├── libfftw3f_threads.a
    ├── libfftw3f_threads.dylib -> libfftw3f_threads.3.dylib
    ├── libfftw3f_threads.la
    └── pkgconfig
        ├── fftw3.pc
        └── fftw3f.pc
```

tkelman:
And what does BinDeps.debug say?

ararslan:
```
The package declares 2 dependencies.
 - Library "libfftw3_threads"
    - Satisfied by:
      - Simple Build Process at /Users/alex/.julia/v0.7/FFTW/deps/usr/lib/libfftw3_threads.3.dylib
    - Providers:
      - BinDeps.AptGet package libfftw3-double3 (can't provide)
      - BinDeps.Pacman package fftw (can't provide)
      - BinDeps.Zypper package libfftw3 (can't provide)
      - BinDeps.Yum package fftw (can't provide)
      - BinDeps.BSDPkg package fftw3 (can't provide)
      - Simple Build Process
 - Library "libfftw3f_threads"
    - Satisfied by:
      - Simple Build Process at /Users/alex/.julia/v0.7/FFTW/deps/usr/lib/libfftw3f_threads.dylib
    - Providers:
      - BinDeps.AptGet package libfftw3-single3 (can't provide)
      - BinDeps.Pacman package fftw (can't provide)
      - BinDeps.Zypper package libfftw3 (can't provide)
      - BinDeps.Yum package fftw (can't provide)
      - BinDeps.BSDPkg package fftw3 (can't provide)
      - Simple Build Process
```

deps/build.jl
```diff
@@ -70,4 +70,8 @@ provides(BuildProcess, (@build_steps begin
     end
 end), [libfftw, libfftwf])
 
-BinDeps.@install Dict([:libfftw => :libfftw, :libfftwf => :libfftwf])
+if is_windows()
+    BinDeps.@install Dict([:libfftw => :libfftw3, :libfftwf => :libfftwf3])
+else
+    BinDeps.@install Dict([:libfftw => :libfftw3_threads, :libfftwf => :libfftwf3_threads])
```
tkelman: I think these are supposed to be the variable name for the library_dependency within this file and the variable name that it should have inside deps.jl, not the value of what that name should be

ararslan: Then I think this should be `:libfftw => :libfftw` as it was before. That also doesn't get the deps.jl file to generate correctly though.

ararslan:
Yeah, as for why this isn't working, I've got nothing so far. Maybe we should call for BinDeps backup: ping @staticfloat

staticfloat:
The reason this isn't working is that[when we check to see if this library [is, in fact, within the BinDeps `load_cache`](https://github.com/JuliaLang/BinDeps.jl/blob/cd6e61aac5eea8c73433d3de39c8f84517e3be99/src/dependencies.jl#L995), we don't find it, because we're looking for `libfftw3f` when we should be looking for `libfftw3f_threads`.   Making [this change](https://github.com/JuliaMath/FFTW.jl/commit/9dabc49784984254093e20d5840728c958d26b7f) works for me.  It's really aggravating that you can't just pass variables in to `BinDeps.@install` because of macro issues.

ararslan:
Hah, I actually had that change earlier but I got the key-value pairs backwards when I did it, so it didn't work. THANK YOU

ararslan:
obsolete, closing in favor of #14

PR #21, opened 2017-07-06
Wrap everything in a version check

ararslan:
The diff looks a little weird, but this just wraps the contents of src/FFTW.jl with
```julia
if VERSION < v"0.7.0-DEV.602"
    import Base.FFTW
else
    # do everything that was here before
end
```


stevengj:
This still defines a `module FFTW; import Base.FFTW; end` in 0.6.

My thinking was that the entire `FFTW.jl` file should just consist of `import Base.FFTW` in 0.6 (via a versioned REQUIRE file ... I don't think it's worth worrying about older versions of 0.7-dev).

giordano:
Yes, I was thinking about branching the package and keeping an (almost) empty package just for Julia 0.6.

stevengj:
I guess you could also do:
```
if VERSION < ....
    import Base.FFTW
else
    __precompile__(true)
    module FFTW
        ....
    end
end
```
if you don't want to mess with multiple branches/versions of REQUIRE.

stevengj:
Or perhaps
```jl
if VERSION < ....
    import Base.FFTW
else
    include("src.jl")
end
```
so that 0.6 doesn't pay the price of parsing the whole `module FFTW ... end`

ararslan:
The fact that wrapping the AbstractFFTs imports was necessary is a little concerning. Perhaps we need to do something similar there but with Base.DFT?

test/runtests.jl
```diff
@@ -2,12 +2,20 @@
 using FFTW
 using Base.Test
 
-# Make super sure we don't pull in Base names
 importall FFTW
-import AbstractFFTs: Plan, fft, ifft, bfft, fft!, ifft!, bfft!,
+if VERSION >= v"0.7.0-DEV.602"
+    import AbstractFFTs: Plan, fft, ifft, bfft, fft!, ifft!, bfft!,
```
stevengj: shouldn't `using AbstractFFTs` be enough here?

test/runtests.jl
```diff
@@ -2,12 +2,20 @@
...
+                         plan_fft, plan_ifft, plan_bfft, plan_fft!, plan_ifft!, plan_bfft!,
+                         rfft, irfft, brfft, plan_rfft, plan_irfft, plan_brfft,
+                         fftshift, ifftshift, plan_inv
+    import FFTW: fftw_vendor
+else
+    import Base.DFT: Plan, fft, ifft, bfft, fft!, ifft!, bfft!,
                      plan_fft, plan_ifft, plan_bfft, plan_fft!, plan_ifft!, plan_bfft!,
```
stevengj: I'm confused, why do you need to import `plan_fft` in 0.6, since that is already exported?   The only thing here that is not exported in 0.6 is `Plan`.

test/runtests.jl
stevengj: (except for `Plan`)

ararslan: No, to avoid name conflicts, AbstractFFTs only exports things if none of the functions are defined in Base. So `using AbstractFFTs` won't import anything into Main until after the deprecation period.

src/package.jl
```diff
@@ -0,0 +1,61 @@
...
+const depsfile = joinpath(dirname(@__DIR__), "deps", "deps.jl")
+if isfile(depsfile)
+    include(depsfile)
+else
+    error("FFTW is not properly installed. Please run Pkg.build(\"FFTW\") ",
+          "and restart Julia.")
+end
+
+# MKL provides its own FFTW
+fftw_vendor() = Base.BLAS.vendor() === :mkl ? :mkl : :fftw
```
stevengj: Shouldn't this function just be called `vendor()`, with a qualified name of `FFTW.vendor()` analogous to `BLAS.vendor()`?

test/runtests.jl
ararslan: I just copied and pasted, but originally I had tried with just `Plan` but there was something else it was getting mad about (can't recall what it was offhand, `plan_` something or other)

src/package.jl
ararslan: Yes. The name is a carryover from Base.

stevengj: with a `@deprecate fftw_vendor` definition, of course.

ararslan: Right

test/runtests.jl
stevengj: That's really annoying.  I feel like we should make it possible to override a deprecated binding without a warning (precisely for things like this … it's going to be a continual annoyance to move things out of Base otherwise)

ararslan: It's the same thing we did with Primes and continue to do with SpecialFunctions, QuadGK, the priority queue stuff in DataStructures, etc. It's certainly not ideal. Being able to do that would make it easier to fully support multiple versions at once.

ararslan: Though AbstractFFTs _does_ `import` and extend the Base definitions, but when they're deprecated and still defined in Base, which sort of simulates exporting its own. That's 0.7.0-DEV.602 and higher though.

stevengj: I thought this was already fixed in Base in https://github.com/JuliaLang/julia/pull/18839/commits/fe3b08121d4ab3d929a1921e3beb0e360d2d329e ....

stevengj: Part of the problem is that because of the way we do the `Base.DFT` deprecation, we never actually call `Base.deprecate(Main, :fft)` etcetera.  This is easy to fix.

stevengj: I just verified that this works:
```
julia> Base.deprecate(Main.DFT, :fft)

julia> include("Foo.jl"); using .Foo

julia> fft([1,2,3])
3-element Array{Int64,1}:
 2
 3
 4
```
where `Foo.jl` is define a module `Foo` that exports `fft(x) = x+1`. 

So, it looks like we just need to add the missing `deprecate` calls to Julia 0.7 and we should be all set.

ararslan: I'm confused. Where does that code need to live? I'm not familiar with using the `deprecate` function directly, though I assume that's what `@deprecate` is calling under the hood.

stevengj: The problem is that we don't use `@deprecate` for the Base.DFT deprecations: we just define an ordinary method and call `error` inside of it.

ararslan: Ohhhh that's right. How else do you get the Base message to point to the package though?

stevengj: We can continue to do what we are doing, we just need to call `deprecate` manually on the symbol, I think.

ararslan: What does that actually do?

stevengj: And actually, it seems that `@deprecate` doesn't call `deprecate` either.... only `@deprecate_binding` uses it.

ararslan: Is it actually legit to deprecate in Main though? What happens to packages then?

stevengj: Oh, sorry, the code above relied on patching Julia as well to check for the deprecation before emitting a warning message:
```
--- a/src/module.c
+++ b/src/module.c
@@ -190,7 +190,7 @@ static jl_binding_t *jl_get_binding_(jl_module_t *m, jl_sym_t *var, modstack_t *
                 if (tempb == NULL || tempb->owner == NULL)
                     // couldn't resolve; try next using (see issue #6105)
                     continue;
-                if (owner != NULL && tempb->owner != b->owner &&
+                if (owner != NULL && tempb->owner != b->owner && !tempb->deprecated &&
                     !(tempb->constp && tempb->value && b->constp && b->value == tempb->value)) {
                     jl_printf(JL_STDERR,
                               "WARNING: both %s and %s export \"%s\"; uses of it in module %s must be qualified\n",
```

stevengj: Hopefully I'll have a PR for Julia shortly.

ararslan: That would be awesome!

timholy:
I tried `Pkg.checkout("FFTW", "aa/ugh-oh-my-god")` (love the branch name!), but that doesn't fix FFTViews, sadly.

timholy:
Can fix it by changing `FFTW.fft` to `Main.FFTW.fft`.

ararslan:
I think that gets back to the fact that neither this package nor AbstractFFTs exports anything when the names are defined in Base...

ararslan:
Since this gets things working on 0.6 I think we should just go ahead and merge this and I'll continue to try to get 0.7 working in separate PRs.

ararslan:
(Actually let's wait until Compat is fixed)

ararslan:
Actually the fix was wrong in Compat so this should be fine.

PR #22, opened 2017-07-17
Stop using ANY

andreasnoack:
It is not a thing anymore

### DIFF
--- a/src/fft.jl
+++ b/src/fft.jl
@@ -541,7 +541,7 @@ for (Tr,Tc,fftw,lib) in ((:Float64,:Complex128,"fftw",libfftw),
         return rFFTWPlan{$Tc,$BACKWARD,inplace,N}(plan, flags, R, X, Y)
     end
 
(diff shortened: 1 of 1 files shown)


ararslan:
It's not equivalent to `Any` though; I thought it required the new `@nospecialize` macro?

andreasnoack:
I'm not sure but I thought this only made a difference for the arguments of the function and not when used as a type parameter like this. We should probably consult the authorities.

giordano:
Unrelated, the CI logs on nightly are showing dozens of messages like those reported at https://github.com/JuliaLang/julia/pull/22763#issuecomment-315641223 for `SpecialFunctions.jl`, see e.g. https://travis-ci.org/JuliaMath/FFTW.jl/jobs/254462935#L785

ararslan:
@giordano Yeah, FFTW is just a hot mess in general right now.

giordano:
I opened a ticket (#23) as a reminder.

JeffBezanson:
You might need `(::Type{r2rFFTWPlan{$Tr,A,inplace,N}} where A)` for this to match the same methods as before.

PR #27, opened 2017-07-20
Reexport symbols from AbstractFFTs

### LINKED ISSUE #26: Reexport AbstractFFTs symbols?
It's kind of unfortunate to have to do `using FFTW, AbstractFFTs` to be able to get all of the FFT functionality you need, and it's less discoverable for users. It seems like it would be more convenient from a usability perspective to just do `@reexport using AbstractFFTs` here. Thoughts, @stevengj?

ararslan:
Closes #26

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,6 @@
 julia 0.6.0
 AbstractFFTs 0.2.0
+Reexport
 Compat 0.27.0
 BinDeps 0.6.0
 @osx Homebrew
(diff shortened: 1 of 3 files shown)


ararslan:
The CI failure is because the push build didn't get retriggered, which means it wasn't rebased from master. The PR build shows that it works as expected.

ararslan:
Whoops, wrong button

PR #30, opened 2017-08-16
Use the FFTW binaries from fftw-builder

ararslan:
Let's see what happens...

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -41,15 +41,15 @@ const URL = "https://github.com/ararslan/fftw-builder/releases/download/v$FFTW_V
 # Mapping of Sys.MACHINE to (url, sha) for precompiled binaries from fftw-builder
 const downloads = Dict(
     "x86_64-pc-linux-gnu" => ("$URL-linux-x86_64.tar.gz",
-                              "f8382542f93b391590faee540d71ed8022d5e04037ae664e8f8a6a445b2ea876"),
+                              "43a049496d47dd1919283a981d21078942d04b067edd75b61cc6aef538098b53"),
     "i686-pc-linux-gnu"   => ("$URL-linux-i686.tar.gz",
-                              "b3a7e6726470de92407259ad198d9d9e256df77a3e0276de5d9a81965f5c5176"),
+                              "0b6ec1440ea76f72ed6e5d105a4d634e195b484e847c521ac0812092069f0c74"),
     "x86_64-apple-darwin" => ("$URL-osx-x86_64.tar.gz",
-                              "97fa4cca555d4587c1d2a02d749c92c513cc59eb209a19d3882009006d911c6b"),
+                              "193f20bba844fd2524f70e941aa5203b153bdcd321df2f20fc3a9d15b4aae50a"),
     "x86_64-w64-mingw32"  => ("$URL-win-x86_64.zip",
-                              "40199c4de73e23c0009acb625b14999130d489621d357c9d88003559089d69eb"),
+                              "4e9bd1022c5980869b06bb9d10767fd2578dfe47f9323e6f572d0918da8c8a07"),
     "i686-w64-mingw32"    => ("$URL-win-i686.zip",
-                              "47eeac9c8a8a869ac0d9114e4a57dc6463aa4dd4e7dd12eaf73aea9993c04b8f"),
+                              "60a65fd689495629a1558858f3a52011f9ce28df8aa756f5b315b17dd8dd8843"),
 )
 
 const machine = Sys.isapple() ? "x86_64-apple-darwin" : Sys.MACHINE
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -2,8 +2,10 @@ using BinDeps
 using BinDeps: builddir, usrdir
 
 # Binaries is not a recognized provider on Linux >:/
-if Sys.islinux()
-    prepend!(BinDeps.defaults, [BinDeps.Binaries])
+modified_defaults = false
+if !in(BinDeps.Binaries, BinDeps.defaults)
+    unshift!(BinDeps.defaults, BinDeps.Binaries)
+    modified_defaults = true
 end
 
 BinDeps.@setup
@@ -55,7 +57,8 @@ const machine = Sys.isapple() ? "x86_64-apple-darwin" : Sys.MACHINE
 if haskey(downloads, machine)
     url, sha = downloads[machine]
     isdir(usrdir(libfftw)) || mkpath(usrdir(libfftw))
-    provides(Binaries, URI(url), [libfftw, libfftwf], SHA=sha, os=BinDeps.OSNAME)
+    provides(Binaries, URI(url), [libfftw, libfftwf], SHA=sha, os=BinDeps.OSNAME,
+             unpacked_dir="fftw-$FFTW_VER")
     scratch = false
 elseif Sys.KERNEL === :FreeBSD
     provides(BSDPkg, "fftw3", [libfftw, libfftwf], os=:FreeBSD)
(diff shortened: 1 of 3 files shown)


ararslan:
At this point, this looks like a failure of the BinDeps setup rather than one of the binaries themselves. Any thoughts or suggestions, @tkelman and/or @staticfloat?

deps/build.jl
```diff
@@ -33,21 +32,34 @@ end
 libfftw = library_dependency(libfftw_name, aliases=makealiases(libfftw_name))
 libfftwf = library_dependency(libfftwf_name, aliases=makealiases(libfftwf_name))
 
-provides(AptGet, "libfftw3-double3", [libfftw], os=:Linux)
-provides(AptGet, "libfftw3-single3", [libfftwf], os=:Linux)
-provides(Pacman, "fftw", [libfftw, libfftwf], os=:Linux)
-provides(Zypper, "libfftw3_threads3", [libfftw, libfftwf], os=:Linux)
-provides(Yum, "fftw-libs", [libfftw, libfftwf], os=:Linux)
-provides(BSDPkg, "fftw3", [libfftw, libfftwf], os=:FreeBSD)
+url = "https://github.com/ararslan/fftw-builder/releases/download/v\$FFTW_VER/libfftw-\$FFTW_VER"
```
musm: I don't think you meant `\$`  -> `$`

musm: nvm I just saw the huge include_string above....

staticfloat: Yeah, this huge `include_string()` is nasty; we should instead just use `include()` and put the newer build instructions in a separate file

deps/build.jl
```diff
@@ -33,21 +32,34 @@ end
...
+    provides(Binaries, URI("\$url-linux-\$(Sys.ARCH).tar.gz"), [libfftw, libfftwf],
+             unpacked_dir="usr/lib", SHA=sha, os=:Linux)
+elseif Sys.iswindows() && Sys.ARCH in [:x86_64, :i686]
+    sha = if Sys.ARCH === :x86_64
+        "e0376b43bcd161762726f683b4404844388fc2b8997ba86405555f844d390f37"
+    elseif Sys.ARCH === :i686
+        "67b8084cfed5f240aff70890369111f38fed624e8fbae58719c3d1ef58ca02ff"
+    end
+    provides(Binaries, URI("\$url-win-\$(Sys.ARCH).zip"), [libfftw, libfftwf],
+             unpacked_dir="usr/lib", SHA=sha, os=:Windows)
```
staticfloat: I would suggest something a little more ordered, like what [Tom did for `Sundials.jl`](https://github.com/JuliaDiffEq/Sundials.jl/blob/7b679d931198cc5c646ca29390ff3ac9879b67cb/deps/build.jl).  Maybe with a check to see if there just aren't any platforms available at all.

staticfloat:
Overall looks reasonable, and the errors you're seeing are, ironically, due to `BinDeps.jl`, not `BinDeps2.jl`.  I don't know why it's crashing with an `InterruptException`.  That's weird.

tkelman:
because it's spawning a process to check if sudo works and that spawning fails

ararslan:
> not BinDeps2.jl

Dunno why they would be coming from BinDeps2 anyway, since that isn't being used here...

deps/actual_build.jl
```diff
@@ -0,0 +1,102 @@
...
+libfftw = library_dependency(libfftw_name, aliases=makealiases(libfftw_name))
+libfftwf = library_dependency(libfftwf_name, aliases=makealiases(libfftwf_name))
+
+isdir(joinpath(@__DIR__, "usr", "lib")) || mkpath(joinpath(@__DIR__, "usr", "lib"))
+
+const URL = "https://github.com/ararslan/fftw-builder/releases/download/v$FFTW_VER/libfftw-$FFTW_VER"
+
+# Mapping of (Sys.KERNEL, Sys.ARCH) to (url, sha) for precompiled binaries from fftw-builder
+const downloads = Dict(
+    (:Linux, :x86_64)   => ("$URL-linux-x86_64.tar.gz",
```
tkelman: This needs to check the full GNU triple of `Sys.MACHINE`, just kernel and arch aren't good enough. If someone is running Julia on a non-glibc Linux distro, these binaries won't work and this should use a source build.

ararslan:
It keeps trying to build from scratch on Linux. It's like it's not even seeing the `provides(Binaries, ...)` bit. :( Even on Windows it says none of the declared providers can satisfy the library requirement.

deps/actual_build.jl
ararslan: Am I correct that the appropriate keys are then:
* `x86_64-pc-linux-gnu`
* `i686-pc-linux-gnu`
* `x86_64-w64-mingw32`
* `i686-w64-mingw32`
* `x86_64-apple-darwin` (+ some numbers at the end)

ararslan: The macOS binaries were built with XCode 8 and _should_ work with other versions of macOS...

tkelman: looks right - there will be more possible varieties that should map to "glibc version of linux that would work" but those are going to need to be collected in a single place

ararslan:
I see the issue: The binaries built on Travis have a directory structure in the tarball, whereas the rest of them just have the binaries right inside. I just need to not do that and it should be good I think.

ararslan:
Well, that's _an_ issue at least. Doesn't explain the Windows failure.

ararslan:
Okay. Summary of current bad things:
* 64-bit Linux still ignores the binaries and builds from scratch
* Windows says `UndefVarError: libfftw not defined` (maybe related to how I set `libfftw_name`?)
* macOS says `tar: could not chdir to '/Users/travis/.julia/v0.7/FFTW/deps/libfftw3_threads'`

Any ideas, BinDeps masters?

ararslan:
Awesome, Windows works now! Linux x64 is still building from scratch (it isn't even hitting the `info` I put in there :/) though.

deps/build.jl
```diff
@@ -30,19 +29,31 @@ end
...
+
+const machine = Sys.isapple() ? "x86_64-apple-darwin" : Sys.MACHINE
+
+if haskey(downloads, machine)
+    url, sha = downloads[machine]
+    provides(Binaries, URI(url), [libfftw, libfftwf], SHA=sha, unpacked_dir=".", os=BinDeps.OSNAME)
+elseif Sys.KERNEL === :FreeBSD
+    provides(BSDPkg, "fftw3", [libfftw, libfftwf], os=:FreeBSD)
+else
+    info("No precompiled binaries found for your system. Building from scratch...")
```
ararslan: @staticfloat Any idea why Linux seems to be building from scratch without ever seeing this line?

staticfloat: Put `info()`'s in the other branches of the `if`, I bet it's passing the `haskey()` part and actually running the `provides(Binaries, ....)` on line 52.

staticfloat: Also see if you can't put a `BinDeps.debug("FFTW")` into the travis script to see if it even sees the `provides(Binaries, ...)`.

ararslan: So apparently it's taking that path with the `haskey` since with the most recent commit it triggers the `info`, but it's still ignoring the `provides(Binaries, )` and using the `provides(Sources, )`.

ararslan: Good idea!

ararslan:
```
$ julia -e 'Pkg.add("BinDeps"); using BinDeps, FFTW; BinDeps.debug("FFTW")'
INFO: No packages to install, update or remove
INFO: Package database updated
INFO: Reading build script...
ERROR: UndefVarError: include not defined
Stacktrace:
 [1] debug_context(::String) at /home/travis/.julia/v0.7/BinDeps/src/debug.jl:54
 [2] debug(::Base.PipeEndpoint, ::String) at /home/travis/.julia/v0.7/BinDeps/src/debug.jl:59
 [3] debug(::String) at /home/travis/.julia/v0.7/BinDeps/src/debug.jl:71
```
wtf?

staticfloat:
Okay, apparently `BinDeps.debug("")` doesn't like `include()`.  That's interesting.

I guess next is to debug this locally, on a docker instance or something.

PR #41, opened 2017-09-29
Mention GPL in the README

### LINKED ISSUE #39: FFTW is GPL. Mention in licence file?!
and probably also README?

I think "GPL applies to dependency" should be prominently added to license file. Before MIT.

Only seeing "MIT license" is misleading, and users may want to know about GPL or get in legal trouble otherwise(?).

I think this also applies to DSP.jl (but not AbstractFFTs.jl).

ararslan:
Fixes #39.

Note: CI skipped, remove from commit message when merging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,12 @@ This package provides Julia bindings to the [FFTW](http://www.fftw.org/) library
 fast Fourier transforms, as well as functionality useful for signal processing.
 These functions were formerly a part of Base Julia.
 
+The FFTW library will be downloaded on versions of Julia where it is no longer distributed
+as part of Julia.
+Note that FFTW is licensed under GPLv2 or higher (see
+[its license file](http://www.fftw.org/doc/License-and-Copyright.html)), but the bindings
+here are licensed under MIT.
+
 [![Travis](https://travis-ci.org/JuliaMath/FFTW.jl.svg?branch=master)](https://travis-ci.org/JuliaMath/FFTW.jl)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,6 +4,12 @@ This package provides Julia bindings to the [FFTW](http://www.fftw.org/) library
 fast Fourier transforms, as well as functionality useful for signal processing.
 These functions were formerly a part of Base Julia.
 
+The FFTW library will be downloaded on versions of Julia where it is no longer distributed
+as part of Julia.
+Note that FFTW is licensed under GPLv2 or higher (see
+[its license file](http://www.fftw.org/doc/License-and-Copyright.html)), but the bindings
+here are licensed under MIT.
```
stevengj: This is a bit unclear.   Best to spell out the consequences:
```
This means that code using the FFTW library via the FFTW.jl bindings is subject to FFTW's licensing terms.  Code using alternative implementations of the FFTW API, such as [MKL's FFTW3 interface](https://software.intel.com/en-us/mkl-developer-reference-c-fftw3-interface-to-intel-math-kernel-library) are instead subject to the alternative's license.
```

ararslan: Ah yes, MKL deserves a mention here, but I'm unclear on what "code using the FFTW library via the FFTW.jl bindings is subject to FFTW's licensing terms" means.

stevengj: If you distribute a derived/combined work, i.e. a program that links to FFTW, then that distribution falls under the terms of the GPL (e.g. must include complete source code etc.)

If you just distribute source code that links to FFTW.jl, and users have to download FFTW or MKL to provide the backend, then the GPL probably doesn't have much effect on you.

PR #42, opened 2017-09-29
Use MKL for FFTs if Julia's BLAS is MKL

### LINKED ISSUE #40: speed regression compared to v0.6
I'm seeing a significant slowdown with FFTW in v0.7 as compared to v0.6. Any idea what is going on? Note: I'm using MKL for my ffts.

Here are the timings for v0.7 .... (I would normally use BenchmarkTools to do the timings but it currently throws an error on v0.7)

```julia
julia> using FFTW

julia> FFTW.set_num_threads(8);

julia> m = rand(Float32,16384,16384);

julia> rfft(m); # warmup

julia> @time rfft(m)
  4.607740 seconds (456 allocations: 1.000 GiB, 0.92% gc time)
```

Here are the timings for v0.6 ....

```julia
julia> using FFTW

julia> FFTW.set_num_threads(8);

julia> m = rand(Float32,16384,16384);

julia> rfft(m); # warmup

julia> @time rfft(m)
  1.473979 seconds (426 allocations: 1.000 GiB, 4.71% gc time)
```

Version info for v0.7

```
julia> versioninfo()
Julia Version 0.7.0-DEV.1957
Commit c3a3a1c89f* (2017-09-25 23:52 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin17.0.0)
  CPU: Intel(R) Core(TM) i7-7920HQ CPU @ 3.10GHz
  WORD_SIZE: 64
  BLAS: libmkl_rt
  LAPACK: libmkl_rt
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, broadwell)
Environment:
  JULIA_NUM_THREADS = 8
```

Version info for v0.6

```
Julia Version 0.6.1-pre.3
Commit 4aa661a0c1* (2017-09-14 17:43 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin17.0.0)
  CPU: Intel(R) Core(TM) i7-7920HQ CPU @ 3.10GHz
  WORD_SIZE: 64
  BLAS: libmkl_rt
  LAPACK: libmkl_rt
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, broadwell)
```

ararslan:
@EthanAnderes, could you clone this branch and try it out? In particular we want to make sure that the file `~/.julia/v0.7/FFTW/deps/deps.jl` exists and says MKL in it, and that you get the expected performance boost.

If all is well, fixes #40.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,120 +1,21 @@
-using BinDeps
-using BinDeps: builddir, depsdir, libdir
-
-# Binaries is not a recognized provider on Linux >:/
-modified_defaults = false
-if !in(BinDeps.Binaries, BinDeps.defaults)
-    unshift!(BinDeps.defaults, BinDeps.Binaries)
-    modified_defaults = true
-end
-
-BinDeps.@setup
-
-const FFTW_VER = v"3.3.6-pl2"
-
-if Sys.iswindows()
-    const libfftw_name = "libfftw3"
-    const libfftwf_name = "libfftw3f"
+# If BLAS was compiled with MKL, we want to use MKL for FFTs as well. Thus
+# we have to do this little dance to get around having to use BinDeps for
+# a library that's already linked to Julia.
+if Base.BLAS.vendor() === :mkl
+    mklpath = Libdl.dlpath("libmkl_rt")
+    depsfile = joinpath(@__DIR__, "deps.jl")
+    isfile(depsfile) && rm(depsfile, force=true)
+    open(depsfile, "w") do f
+        println(f, """
+            # This is an auto-generated file, do not edit
+            if Libdl.dlopen_e("$mklpath") == C_NULL
+                error("Unable to load MKL from '$mklpath'.\\n",
+                      "Please rerun Pkg.build(\\"FFTW\\") and restart Julia.")
+            end
+            const libfftw = "$mklpath"
+            const libfftwf = "$mklpath"
+        """)
+    end
 else
-    const libfftw_name = "libfftw3_threads"
-    const libfftwf_name = "libfftw3f_threads"
-end
-
-# Why can't everyone just agree on what to call this library...
-function makealiases(lib)
-    major = string(FFTW_VER.major)
-    nover = replace(lib, major, "")
-    return String[
-        nover,
-        join([lib, Libdl.dlext, major], "."),
-        join([nover, Libdl.dlext, major], "."),
-        lib * "-" * major,
-        nover * "-" * major,
-    ]
-end
-
-libfftw = library_dependency(libfftw_name, aliases=makealiases(libfftw_name))
-libfftwf = library_dependency(libfftwf_name, aliases=makealiases(libfftwf_name))
-
-const URL = "https://github.com/ararslan/fftw-builder/releases/download/v$FFTW_VER/libfftw-$FFTW_VER"
-
-# Mapping of Sys.MACHINE to (url, sha) for precompiled binaries from fftw-builder
-const downloads = Dict(
-    "x86_64-pc-linux-gnu" => ("$URL-linux-x86_64.tar.gz",
-                              "43a049496d47dd1919283a981d21078942d04b067edd75b61cc6aef538098b53"),
-    "i686-pc-linux-gnu"   => ("$URL-linux-i686.tar.gz",
-                              "0b6ec1440ea76f72ed6e5d105a4d634e195b484e847c521ac0812092069f0c74"),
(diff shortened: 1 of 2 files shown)


EthanAnderes:
@ararslan Thanks again for acting on this so quick. 

So ... I checked out the branch like this:

```
julia> Pkg.checkout("FFTW", "aa/mkl")
INFO: Checking out FFTW aa/mkl...
INFO: Pulling FFTW latest aa/mkl...
INFO: No packages to install, update or remove
```

... then restarted Julia and didn't see the performance improvement. *However*, when I manually went into `~/.julia/v0.7/FFTW/deps/deps.jl` and edited it as follows ... it worked! Got the speed up. 

```
# Load dependencies
# ---- I manually commented this out ----
# @checked_lib libfftwf "/Users/ethananderes/.julia/v0.7/FFTW/deps/usr/lib/libfftw3f_threads.dylib"
# @checked_lib libfftw "/Users/ethananderes/.julia/v0.7/FFTW/deps/usr/lib/libfftw3_threads.dylib"

# ---- and added this ----
@checked_lib libfftwf "libmkl_rt"
@checked_lib libfftw "libmkl_rt"

```

So it must be a problem with how it is assigns `libfftwf` and `libfftw`. Note my comment I just made on the corresponding issue where `FFTW` was able to get `FFTW.libfftwf_name` and `FFTW.libfftw_name` correctly pointing to MKL, just not `libfftwf` and `libfftw`). 

Hope this helps and sorry I'm not experienced enough to help further. Let me know if there is anything else I can do to help you pin it down.

ararslan:
Did you rerun `Pkg.build("FFTW")` between checking out my branch and rerunning the benchmarks? Rebuilding should rewrite the deps file accordingly.

EthanAnderes:
Ahh, nope, I didn't run `Pkg.build("FFTW")`. Just did it now and it *worked*!
Thanks a ton. Let me know if there is any other testing you want me to do.

 For reference, here is what `~/.julia/v0.7/FFTW/deps/deps.jl` looks like after I ran the build command:

```
    # This is an auto-generated file, do not edit
    if Libdl.dlopen_e("/opt/intel/compilers_and_libraries_2018.0.104/mac/mkl/lib/libmkl_rt.dylib") == C_NULL
        error("Unable to load MKL from '/opt/intel/compilers_and_libraries_2018.0.104/mac/mkl/lib/libmkl_rt.dylib'.\n",
              "Please rerun Pkg.build(\"FFTW\") and restart Julia.")
    end
    const libfftw = "/opt/intel/compilers_and_libraries_2018.0.104/mac/mkl/lib/libmkl_rt.dylib"
    const libfftwf = "/opt/intel/compilers_and_libraries_2018.0.104/mac/mkl/lib/libmkl_rt.dylib"

```

ararslan:
If checking out this branch did it for you, then I think that's all the testing I need. Thanks for your help, much appreciated! Once this PR is merged and a new tag of the package is made, you should be able to do `Pkg.free("FFTW")` to go back to the latest tagged version.

stevengj:
There should be a way to override this, if the user wants features/transforms in FFTW that are not supported by MKL

ararslan:
I guess what we could do when MKL is installed is also install FFTW, then do something like
```
const libfftw = "path/to/libfftw.so"
const mkl_or_fftw = # path to MKL if installed, path to FFTW otherwise
```
then use `mkl_or_fftw` for everything available in both APIs and `libfftw` for things specific to FFTW. I'm not familiar with what the differences are between MKL and FFTW though; I assumed they provided all of the same things.

stevengj:
No, MKL only implements a subset of FFTW's functionality:

https://software.intel.com/en-us/mkl-developer-reference-c-using-fftw3-wrappers

For example, since MKL supports at most 7-dimensional transforms (IIRC), it can't be used for Hadamard.jl

ararslan:
Okay. Does What I said then with the two separate library path constants seem like a good idea then?

stevengj:
It's not sufficient to have two library paths. You'll also need to have two kinds of Plan, because you can't mix the APIs: e.g. a plan created by FFTW can't be executed by MKL. 

Seems tricky. 

I was just thinking of a build-time option to select one or the other.

ararslan:
Ah, I see. I'm not sure that there's an easy way to make a build-time option, as we're kind of limited by `Pkg.build`, which does not accept options.

stevengj:
The usual way is with environment variables. See e.g. PyCall

ararslan:
https://github.com/JuliaMath/FFTW.jl/pull/44

PR #44, opened 2017-10-01
Use an environment variable to request MKL

ararslan:


deps/build.jl
```diff
@@ -1,7 +1,7 @@
-# If BLAS was compiled with MKL, we want to use MKL for FFTs as well. Thus
-# we have to do this little dance to get around having to use BinDeps for
-# a library that's already linked to Julia.
-if Base.BLAS.vendor() === :mkl
+# If BLAS was compiled with MKL and the user wants MKL-based FFTs, we'll oblige.
+# In that case, we have to do this little dance to get around having to use BinDeps
+# for a library that's already linked to Julia.
+if get(ENV, "JULIA_FFTW_PROVIDER") == "MKL" && Base.BLAS.vendor() === :mkl
```
stevengj: The problem with this is that if you do a `Pkg.update()` it will "forget" your previous setting unless the environment variable is still set.  That's why in PyCall I save the setting in a file.

ararslan: Seems like the kind of thing that could go in a .juliarc.jl file

stevengj: Requiring users to not only set an env var, but to set it "permanently" in a file somewhere, seems awfully error prone.

Besides, environment variables live in a global namespace, e.g. they "leak" into Julia subprocesses, so I greatly prefer setups where you only need to set them temporarily.

ararslan:
Okay, now it stores the setting after the first build.

deps/build.jl
```diff
@@ -1,7 +1,16 @@
...
+settings = joinpath(@__DIR__, "..", ".build_settings")
+if isfile(settings)
+    provider = readchomp(settings)
+elseif haskey(ENV, "JULIA_FFTW_PROVIDER")
+    provider = ENV["JULIA_FFTW_PROVIDER"]
+    open(f -> println(f, provider), settings, "w")
+else
+    provider = ""
+end
+if provider == "MKL" && Base.BLAS.vendor() === :mkl
```
stevengj: Shouldn't it be an error if `provider == "MKL"` and MKL is not available?

ararslan: Yeah that's probably better

deps/build.jl
```diff
@@ -1,7 +1,19 @@
...
-if Base.BLAS.vendor() === :mkl
+# If BLAS was compiled with MKL and the user wants MKL-based FFTs, we'll oblige.
+# In that case, we have to do this little dance to get around having to use BinDeps
+# for a library that's already linked to Julia.
+settings = joinpath(@__DIR__, "..", ".build_settings")
+action = ""
+if isfile(settings)
+    provider = readchomp(settings)
+    action = "delete " * normpath(settings)
+elseif haskey(ENV, "JULIA_FFTW_PROVIDER")
```
stevengj: These if statements seem to be in the wrong order to me — the environment variable should override the hidden settings file

deps/build.jl
```diff
@@ -16,6 +28,9 @@ if Base.BLAS.vendor() === :mkl
             const libfftwf = "$mklpath"
         """)
     end
+elseif provider == "MKL"
+    error("MKL build requested for FFTW but Julia was not built with MKL.\n",
+          "To fix this, " action, " and rerun Pkg.build(\"FFTW\").")
```
stevengj: With my above suggestion, you just need to set `ENV["JULIA_FFTW_PROVIDER"]="FFTW"` and it will work.

stevengj:
(I think it's really important for Pkg3 to standardize these kinds of persistent build-time options: e.g.  JuliaLang/Juleps#38 is one rough draft of a possibility)

ararslan:
Squashed my previous commits and added an unrelated commit that should get around the failures on nightly.

Issue #52, opened 2018-01-03
Provide `rfft!` and `irfft!` through a submodule


favba:
Is it desirable to have a submodule living in FFTW.jl to provide the inplace real-to-complex and complex-to-real transformations?

The functionality is already coded in [InpalceRealFFT.jl](https://github.com/favba/InplaceRealFFT.jl) and could remain living as a separate package. Some people [expressed the desire](https://discourse.julialang.org/t/ann-inplacerealfft-jl-inplace-real-to-complex-and-complex-to-real-fft/8116/2) to have it in FFTW.jl
Basically, a `PaddedArray` type is created to deal with proper padding of the real data, and methods  `rfft!`,`irff!`,`brfft!`,`plan_rfft!`,`plan_irfft!` and `plan_brfft!` are created to work with `AbstractPaddedArray`s, an interface to provide custom PaddedArray types.

I could make a PR if this is desirable.

stevengj:
A PR would be good, but don't bother with a submodule.  Just put it directly into the `FFTW` module (in a separate file via `include`) and export `rfft!` etcetera.

Maybe `PaddedRFFTArray` to make it clear that the padding is specific to the semantics of FFTW's in-place `rfft`.

Marc-3d:
Hi, 

I too ended up implementing my own in-place r2c/c2r  (although more naively than favba's implementation). However, I realize now the importance of the thread-safety mechanism of the ```@exclusive``` macro in FFTW, and I started thinking about how to add ```rfft!``` capabilities to FFTW. I had a look at the PR and the discussion around this topic, and I have a suggestion. 

IMO the simplest way to add ```r2c!``` and  ```c2r!``` capabilities would be add two new signatures for ```rFFTWPlan``` with a single input array + an extra parameter about the intended size of the FFT. One could add an ```@assert``` or ```@warning``` statement to make the functions more user-friendly-ish. 

```julia 
@eval @exclusive function rFFTWPlan{$Tr,$FORWARD,true,N}( X::StridedArray{$Tr,N}, fftsize::Dims{N}, region, flags::Integer, timelimit::Real) where {N} 
    @assert size(X,1) == fftsize[1] + 2*iseven( fftsize[1] ) "Informative message about padding requirements for inplace rFFTs"
    @warn "make sure you know what you are doing"
    #= function body =# 
end
```

```julia 
[...]

PR #54, opened 2018-01-05
[W.I.P.] Added rfft! and irfft! functionality through PaddedRFFTArray type.

favba:
This PR addresses #52. This is a rewrite of [my code](https://github.com/favba/InplaceRealFFT.jl) with fixes to address possible problems with usage of `unsafe_wrap` raised [here](https://discourse.julialang.org/t/ann-inplacerealfft-jl-inplace-real-to-complex-and-complex-to-real-fft/8116/13?u=favba). I don't fully understand those issues, so an expert review is needed to see if the changes I introduced indeed solved the problem.

The code is already functional. The test I wrote originally for my package are passing. The code still needs documentation of the new type and new functions.

### DIFF
--- a/src/rfft!.jl
+++ b/src/rfft!.jl
@@ -1,0 +1,216 @@
+import Base: IndexStyle, getindex, setindex!, eltype, \, similar, copy, real, read!
+
+#For compatibility between Julia v0.6 and v0.7 - begin
+if VERSION >= v"0.7-"
+    using Base.@gc_preserve
+else
+    macro gc_preserve(s::Symbol,ex::Expr)
+        return esc(ex)
+    end
+end
+#For compatibility between Julia v0.6 and v0.7 - end
+
+export PaddedRFFTArray, plan_rfft!, rfft!, plan_irfft!, plan_brfft!, brfft!, irfft!
+
+
+
+# As the time this code was written the new `ReinterpretArray` introduced in
+# Julia v0.7 had major performace issues. Those issues were bypassed with the usage of the
+# unsafe_wrap for the complex view of the data. As the name sugest, this is
+# unsafe: whenever this complex view is called one must be careful to gc_preserve
+# the "parent" PaddedRFFTArray. Therefore this view should not be "exported" to the
+# user and the PaddedRFFTArray itself behaves like the complex view.
+# Since it is not possible to not export a particular field of a exported type,
+# a "hack" was used to name the unsafe field "#c", a fieldname that a non-advanced user
+# will likely not be able to call.
+
+const c = Symbol("#c") 
+@eval struct PaddedRFFTArray{T<:fftwReal,N,L} <: DenseArray{Complex{T},N}
+    r::SubArray{T,N,Array{T,N},NTuple{N,UnitRange{Int}},L} # Real view skipping padding
+    ($c)::Array{Complex{T},N}
+
+    function PaddedRFFTArray{T,N}(rr::Array{T,N},nx::Int) where {T<:fftwReal,N}
+        rrsize = size(rr)
+        fsize = rrsize[1]
+        iseven(fsize) || throw(
+            ArgumentError("First dimension of allocated array must have even number of elements"))
+        (nx == fsize-2 || nx == fsize-1) || throw(
+            ArgumentError("Number of elements on the first dimension of array must be either 1 or 2 less than the number of elements on the first dimension of the allocated array"))
+        fsize = fsize÷2
+        csize = (fsize, rrsize[2:end]...)
+        if VERSION >= v"0.7-" 
+            @gc_preserve rr c = unsafe_wrap(Array{Complex{T},N}, 
+                                           reinterpret(Ptr{Complex{T}},pointer(rr)), 
+                                           csize)
+        else 
+            c = reinterpret(Complex{T}, rr, csize)
+        end
+        rsize = (nx,rrsize[2:end]...)
+        r = view(rr,(1:l for l in rsize)...)
+        return  @gc_preserve rr new{T, N, N === 1 ? true : false}(r,c)
+    end # function
+end # struct
+
+@inline real(S::PaddedRFFTArray) = S.r
+
+@inline unsafe_complex_view(S::PaddedRFFTArray) = getfield(S,c)
+
+copy(S::PaddedRFFTArray) = PaddedRFFTArray(copy(parent(real(S))),size(real(S))[1])
+
+similar(f::PaddedRFFTArray,::Type{T},dims::Tuple{Vararg{Int,N}}) where {T, N} =
+    PaddedRFFTArray{T}(dims) 
+similar(f::PaddedRFFTArray{T,N,L},dims::NTuple{N2,Int}) where {T,N,L,N2} =
+    PaddedRFFTArray{T}(dims) 
+similar(f::PaddedRFFTArray,::Type{T}) where {T} =
+    PaddedRFFTArray{T}(size(real(f))) 
+similar(f::PaddedRFFTArray{T,N}) where {T,N} = 
+    PaddedRFFTArray{T,N}(similar(parent(real(f))), size(real(f))[1]) 
+
+size(S::PaddedRFFTArray) =
+    @gc_preserve S size(unsafe_complex_view(S))
+
+IndexStyle(::Type{T}) where {T<:PaddedRFFTArray} = 
+    IndexLinear()
+
+Base.@propagate_inbounds @inline getindex(S::PaddedRFFTArray, i::Int) =
+    @gc_preserve S getindex(unsafe_complex_view(S),i)
+
+Base.@propagate_inbounds @inline getindex(S::PaddedRFFTArray{T,N}, I::Vararg{Int, N}) where {T,N} =
+    @gc_preserve S getindex(unsafe_complex_view(S),I...)
+
+Base.@propagate_inbounds @inline setindex!(S::PaddedRFFTArray,v,i::Int) = 
+    @gc_preserve S setindex!(unsafe_complex_view(S),v,i)
+
+Base.@propagate_inbounds @inline setindex!(S::PaddedRFFTArray{T,N},v,I::Vararg{Int,N}) where {T,N} =
+    @gc_preserve S setindex!(unsafe_complex_view(S),v,I...)
+
+
+PaddedRFFTArray(rr::Array{T,N},nx::Int) where {T<:fftwReal,N} = PaddedRFFTArray{T,N}(rr,nx)
+
+function PaddedRFFTArray{T}(ndims::Vararg{Integer,N}) where {T,N}
+    fsize = (ndims[1]÷2 + 1)*2
+    a = Array{T,N}((fsize, ndims[2:end]...))
+    PaddedRFFTArray{T,N}(a, ndims[1])
+end
+
+PaddedRFFTArray{T}(ndims::NTuple{N,Integer}) where {T,N} =
+    PaddedRFFTArray{T}(ndims...)
+    
+PaddedRFFTArray(ndims::Vararg{Integer,N}) where N = 
+    PaddedRFFTArray{Float64}(ndims...)
+    
+PaddedRFFTArray(ndims::NTuple{N,Integer}) where N = 
+    PaddedRFFTArray{Float64}(ndims...)
+
+function PaddedRFFTArray{T}(a::AbstractArray{<:Real,N}) where {T<:fftwReal,N}
+    t = PaddedRFFTArray{T}(size(a))
+    @inbounds copy!(t.r, a) 
+    return t
+end
+
+PaddedRFFTArray(a::AbstractArray{<:Real}) = PaddedRFFTArray{Float64}(a)
+
+function PaddedRFFTArray(stream, dims)
+    field = PaddedRFFTArray(dims)
+    return read!(stream,field)
+end
+
+function PaddedRFFTArray{T}(stream, dims) where T
+    field = PaddedRFFTArray{T}(dims)
+    return read!(stream,field)
+end
+
+function read!(file::AbstractString, field::PaddedRFFTArray)
+    open(file) do io 
+       return read!(io,field) 
+    end
+end
+
+# Read a binary file of an unpaded array directly to a PaddedRFFT array, without the need
+# of the creation of a intermediary Array. If the data is already padded then the user
+# should just use PaddedRFFTArray{T}(read("file",unpaddeddim),nx)
+function read!(stream::IO, field::PaddedRFFTArray{T,N,L}) where {T,N,L}
+    rr = parent(field.r)
+    dims = size(real(field))
+    nx = dims[1]
+    nb = sizeof(T)*nx
+    npencils = prod(dims)÷nx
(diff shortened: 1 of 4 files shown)


favba:
Due to the rapid changes on Julia now, I'll wait for the release of v0.7 to finish up this PR.

appleparan:
Bump! Could you finish this up with v0.7-dev? It would be nice if this will release before v0.7 official release.

favba:
I think all to major changes on Julia have been made. I'll work on it again.

favba:
I managed to avoid using unsafe views and this PR is ready for technical review now. If it is acceptable, it will only be missing docstrings.
FYI: Using the new `reinterpret` still provides really poor performance:
```julia
julia> versioninfo()
Julia Version 0.7.0-DEV.4702
Commit 524710f60d (2018-03-26 19:44 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin17.4.0)
  CPU: Intel(R) Core(TM) i5-5287U CPU @ 2.90GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, broadwell)
Environment:
  JULIA_NUM_THREADS = 2

julia> using FFTW

julia> a = PaddedRFFTArray(rand(8,8))
5×8 PaddedRFFTArray{Float64,2,false}:
 0.712979+0.28724im   0.509384+0.0177532im  …   0.958592+0.884517im   0.685513+0.401894im
 0.361746+0.369876im  0.968775+0.193823im      0.0521561+0.815999im   0.924453+0.528454im
 0.835827+0.403952im  0.775697+0.187061im       0.106656+0.0442893im  0.321887+0.859492im
 0.784735+0.141277im  0.436122+0.198054im       0.580609+0.22239im    0.266962+0.252452im
      0.0+0.0im            0.0+0.0im                 0.0+0.0im             0.0+0.0im

julia> rfft!(a)
5×8 PaddedRFFTArray{Float64,2,false}:
  32.2515+0.0im       -2.36603+0.797907im  …     2.8308-1.49447im     -2.36603-0.797907im
 0.521591+1.17158im  -0.190847-0.681372im     -0.334622+1.19408im     -2.46524-3.4627im
  1.14537-2.0369im    0.383983-0.265645im      -1.52369-1.83912im    0.0772888+2.13619im
  1.39388-1.77041im    1.46133+3.4213im        -1.50413+0.570226im  0.00780449-0.57811im
   2.6889+0.0im        3.28556-0.875104im       1.79742+1.84643im      3.28556+0.875104im

julia> b = Array(a)
5×8 Array{Complex{Float64},2}:
  32.2515+0.0im       -2.36603+0.797907im  …     2.8308-1.49447im     -2.36603-0.797907im
 0.521591+1.17158im  -0.190847-0.681372im     -0.334622+1.19408im     -2.46524-3.4627im
  1.14537-2.0369im    0.383983-0.265645im      -1.52369-1.83912im    0.0772888+2.13619im
  1.39388-1.77041im    1.46133+3.4213im        -1.50413+0.570226im  0.00780449-0.57811im
   2.6889+0.0im        3.28556-0.875104im       1.79742+1.84643im      3.28556+0.875104im

julia> using BenchmarkTools

julia> @benchmark @inbounds getindex($(b),1)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     1.542 ns (0.00% GC)
  median time:      1.549 ns (0.00% GC)
  mean time:        1.552 ns (0.00% GC)
  maximum time:     5.031 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> @benchmark @inbounds getindex($(a.c),1)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     15.493 ns (0.00% GC)
  median time:      15.699 ns (0.00% GC)
  mean time:        15.729 ns (0.00% GC)
  maximum time:     50.897 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     998

julia> @benchmark @inbounds getindex($(a),1)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     1.846 ns (0.00% GC)
  median time:      1.853 ns (0.00% GC)
  mean time:        1.856 ns (0.00% GC)
  maximum time:     9.709 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> @benchmark sum($b)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     34.567 ns (0.00% GC)
  median time:      34.638 ns (0.00% GC)
  mean time:        34.957 ns (0.00% GC)
  maximum time:     91.601 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     993

julia> @benchmark sum($(a.c))
BenchmarkTools.Trial:
  memory estimate:  208 bytes
  allocs estimate:  6
  --------------
  minimum time:     865.900 ns (0.00% GC)
  median time:      875.400 ns (0.00% GC)
  mean time:        1.011 μs (9.66% GC)
  maximum time:     705.103 μs (99.77% GC)
  --------------
  samples:          10000
  evals/sample:     60

julia> @benchmark sum($(a))
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     35.123 ns (0.00% GC)
  median time:      35.186 ns (0.00% GC)
  mean time:        35.756 ns (0.00% GC)
  maximum time:     70.551 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     992

**Comment by favba** [comment]
Did I mess this pull request up with incorrect rebase/merges stuff? Or this is the way to go? Should I start a new clean PR with those final changes?

**Comment by favba** [comment]
I consider this finished now. Julia v0.7 probably won't have any other disruptive change.
If this PR is accepted I could (slowly) start working on documentation...

File: src/rfft!.jl [hunk 1]
```diff
@@ -0,0 +1,234 @@
...
+    ip = :(2*I2[1])
+    t = Expr(:tuple)
+    for i=2:N
+        push!(t.args,:(I2[$i]))
+    end
+    quote
+        d = data(A)
+        i = $ip
+        I = $t
+        @boundscheck checkbounds(d,i,I...)
```
stevengj: Shouldn't it also check bounds with `i-1`?

stevengj: Oh, I guess since `ip=2*I2[1]`, if it is in-bounds then `i-1` must also be in-bounds.

src/rfft!.jl
```diff
@@ -0,0 +1,234 @@
...
+##########################################################################################
+# Inverse plans
+
+function plan_brfft!(X::PaddedRFFTArray{T,N}, region;
+                    flags::Integer=PRESERVE_INPUT,
+                    timelimit::Real=NO_TIMELIMIT) where {T<:fftwReal,N}
+    (1 in region) || throw(ArgumentError("The first dimension must always be transformed"))
+    if flags&PRESERVE_INPUT != 0
+        a = similar(X)
+        return rFFTWPlan{Complex{T},BACKWARD,true,N}(complex_view(a), real(a), region, flags,timelimit)
```
stevengj: What's going on here?   IIRC, `PRESERVE_INPUT` only  affects out-of-place plans.

favba: Oh my... That is reminiscent from early stage development, when I didn't noticed that inplace r2c and c2r were already exposed on Julia. I tried hacking out-of-place plans to work inplace, thus this PRESERVE_INPUT flag.
I'll change this.

favba:
Sorry for this late change. But I just observed that performance is slightly better if we use `Colon`s instead of `UnitRage`s for the array view:
```julia
julia> a = rand(128,128,128);

julia> b = view(a,1:126,:,:);

julia> c = view(a,1:126,1:128,1:128);

julia> using BenchmarkTools

julia> @benchmark sum($b)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     4.533 ms (0.00% GC)
  median time:      4.620 ms (0.00% GC)
  mean time:        4.689 ms (0.00% GC)
  maximum time:     6.707 ms (0.00% GC)
  --------------
  samples:          1064
  evals/sample:     1

julia> @benchmark sum($c)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     5.150 ms (0.00% GC)
  median time:      5.208 ms (0.00% GC)
  mean time:        5.267 ms (0.00% GC)
  maximum time:     7.113 ms (0.00% GC)
  --------------
  samples:          947
  evals/sample:     1

julia> @benchmark reduce($(*),$b)
BenchmarkTools.Trial:
  memory estimate:  560 bytes
  allocs estimate:  14
  --------------
  minimum time:     4.531 ms (0.00% GC)
  median time:      4.608 ms (0.00% GC)
  mean time:        4.678 ms (0.00% GC)
  maximum time:     6.404 ms (0.00% GC)
  --------------
  samples:          1067
  evals/sample:     1

julia> @benchmark reduce($(*),$c)
BenchmarkTools.Trial:
  memory estimate:  560 bytes
  allocs estimate:  14
  --------------
  minimum time:     5.161 ms (0.00% GC)
  median time:      5.227 ms (0.00% GC)
  mean time:        5.291 ms (0.00% GC)
  maximum time:     7.780 ms (0.00% GC)
  --------------
  samples:          943
  evals/sample:     1

julia> versioninfo()
Julia Version 0.7.0-beta2.0
Commit b145832402* (2018-07-13 19:54 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin17.6.0)
  CPU: Intel(R) Core(TM) i5-5287U CPU @ 2.90GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.0 (ORCJIT, broadwell)
Environment:
  JULIA_NUM_THREADS = 2
```

src/rfft!.jl
```diff
@@ -9,27 +9,27 @@ export PaddedRFFTArray, plan_rfft!, rfft!, plan_irfft!, plan_brfft!, brfft!, irf
...
             ArgumentError("First dimension of allocated array must have even number of elements"))
         (nx == fsize-2 || nx == fsize-1) || throw(
             ArgumentError("Number of elements on the first dimension of array must be either 1 or 2 less than the number of elements on the first dimension of the allocated array"))
-        fsize = fsize÷2
-        csize = (fsize, rrsize[2:end]...)
         c = reinterpret(Complex{T}, rr)
-        rsize = (nx,rrsize[2:end]...)
-        r = view(rr,(1:l for l in rsize)...)
-        return  new{T, N, N === 1 ? true : false}(rr,r,c)
+        r = view(rr, 1:nx, ntuple(i->Colon(),Nm1)...)
```
stevengj: You can pass `Val{Nm1}()` here to get a compiler-unrolled `ntuple` on 0.7.   On 0.6 you need `Val{Nm1}`.

favba: Because `Val(n)` is a pure function now I can just put `Val(Nm1)` instead of `Val{Nm1}()`, right?

stevengj:
LGTM, now just needs docs.

favba:
Just to give some update. I' currently finishing my thesis, so I don't think I'll be able to work on this until a month from...

navidcy:
@favba news on this PR? Seems neat and it would be very nice to have it merged in FFTW.jl

roflmaostc:
I would also very appreciate that feature!

BioTurboNick:
I took a stab at reviving this, but it has some issues. My motivation was to allow simply replacing `irfft` with `irfft!` in code like `ImageFiltering.jl`, where the new type can be used internally to manage the memory. `rfft!` is the trickier one, given that the output array needs to be bigger than the input array.

There's little difference in performance because most of the time is spent in the C routines, but there is less memory churn. My version can be found here: https://github.com/BioTurboNick/FFTW.jl/tree/in-place-rfft

I don't know the API that well, so I've hit a stumbling block in a couple places. Maybe someone else can pick up from here.

BioTurboNick:
A few issues:

- Not able to use `irfft!` directly on the output of `rfft`
- `real` function should probably be renamed, as it is doing something quite different from the `real` function in Base.
- An inconsistency: the `irfft!` and `brfft!` functions here allow the form `irfft!(b, (1, 3))` but the existing functions require an integer to be provided e.g. `irfft(b, 8, (1, 3))`

BioTurboNick:
Also, not sure if all the special indexing methods in this PR are necessary anymore, as those are noted as needed due to inefficiencies back in Julia v0.7.

favba:
> Does my PR fail in some way? Because it implements what I suggested with respect to being able to use `irfft!` on `rfft` output. That is, without changing the return type of `rfft`.

Ok, I see it now... Since you kept the `irfft` signature, which requires explicitly the first dimension of the real array, you are able to provide that feature...

I think that can be easily incorporated into this PR, keeping different signatures for `irfft!` when it is applied to a `PaddedRFFTArray` or an `Array{Complex{T}}`.

And I think I know a way to make `rfft!` work at best effort at least with `SubArray`s such that the following would work:
```julia
a = rand(256,256)
b = rfft!(irfft!(rfft(a),size(a,1))) # same as rfft(a)
```

@BioTurboNick , do you mind If I incorporate those features you provided in your PR to this one? or would you rather keep working on that one?
The most important features, at least for me, already implemented on this PR is the efficient indexing and the capability of reading data directly to a `PaddedRFFTArray`, such that absolutely no copy of data is necessary. (I've been using it to perform FFTs of real data that hardly fits memory...)

favba:
Maintaining good index performance while adding those features is trickier than I thought it would be...
I think I'll work on that on a different PR, as it will involve finding a different solution for the slow `reinterpret` problem.

BioTurboNick:
Ah, dang. Fair enough. Wonder what would be necessary to resolve the underlying slowness... I have a tendency to poke around on problems like that.

favba:
> Try `reinterpret(reshape, T, A)`. I've seen this be 20x faster than `reshape(reinterpret(T, A), sz)`. Requires Julia 1.6+

I think the issue is at a lower level than that... see my benchmark [here](https://github.com/JuliaMath/FFTW.jl/pull/54#issuecomment-939341627). It doesn't seem to involve any reshaping


> I was about to suggest `unsafe_wrap` but you already moved away from that... But looking at the concerns around it, maybe there's a way to create the PaddedRFFTArray type to address these issues?

That is my plan... to fiddle with references and/or pointers again.
Basically, my idea is to change the `data` field from an `Array{T,N}` to something like `WrappedArray{T,N,A}` where `WrappedArray` wraps either a real-valued or a complex-valued array (`A` being the type of the array), depending on how you initialize it, but indexing it always return a real value.

I'd put the "unsafe" indexing routines inside this WrappedArray (in the case it is initialized with a complex array), so it is hidden in this extra layer, and maybe keep the current indexing strategy for the complex view of the array.
This `WrappedArray` would also be useful to initialize the `PaddedRFFTArray` directly with the complex data so
```julia
a = rand(256,256)
irff!(rfft(a),256)
```
should work.

favba:
The new implementation is [here](https://github.com/JuliaMath/FFTW.jl/pull/222).

PR #74, opened 2018-06-05
switch to BinaryBuilder binaries

### LINKED ISSUE #73: Migrate the build to BinaryBuilder
Is there any effort to migrating the build to BinaryBuilder?

stevengj:
Closes #73.


stevengj:
@ararslan, is there test coverage for the MKL version?

ararslan:
No, I don't think MKL is available on any CI providers.

ararslan:
I don't, unfortunately.

andreasnoack:
> No, I don't think MKL is available on any CI providers.

https://github.com/travis-ci/travis-ci/issues/5381#issuecomment-281983826 might be relevant here

stevengj:
To use this MKL, it seems like the build script should be updated so that you can simply specify the path of an MKL library rather than the one included with Julia...

stevengj:
Another possibility that occurs to me is to use the `mkl_fft` Anaconda package with Conda.jl.

stevengj:
Updated the PR to add a Conda dependency, so that we can support using MKL even if Julia was not built with MKL support.  Now Travis can exercise the MKL API.

REQUIRE
```diff
@@ -3,3 +3,4 @@ AbstractFFTs 0.3.0
 Reexport
 Compat 0.62.0
 BinaryProvider 0.3.0
+Conda
```
ararslan: Unfortunate to have to add a dependency on Conda but I don't know of a way around it...

stevengj: Note that the Conda dependency is actually quite small if you aren't using it — it only actually runs the Miniconda installer when you decide to add a Conda package.

stevengj:
Okay to merge?

([As I wrote above](https://github.com/JuliaMath/FFTW.jl/pull/74#discussion_r194522518), the Conda.jl dependency is lightweight because Miniconda is only installed lazily.)

stevengj:
Intel provides Conda packages via https://github.com/IntelPython/

EthanAnderes:
I ran into a small hiccup trying this out. However, I'm only an avid user so hope I'm not adding noise to the dev process.

EthanAnderes:
just realized that perhaps this is due to `dev FFTW` + switching to a different branch in the terminal, not automatically installing new additions in REQUIRE. Sorry for the noise.

stevengj:
@EthanAnderes, yes, that is JuliaLang/Pkg.jl#367

stevengj:
@ararslan, okay to merge?

ararslan:
Yep, looks good to me. Feel free to merge or squash-merge at your leisure.

PR #87, opened 2019-01-29
Add missing `undef` to array allocation

darwindarak:


giordano:
I guess this function isn't covered by tests. It would probably be useful to add an appropriate test.

stevengj:
Ping.  This should be merged, but it would be nice to fix test coverage as well.

stevengj:
I'll merge this since it is clearly a bugfix, but a PR with a testcase would be appreciated.

PR #88, opened 2019-02-04
Add Cirrus script to test the package on FreeBSD

giordano:
If someone with admin rights on this repo con [install Cirrus](https://github.com/marketplace/cirrus-ci), this PR enables testing for FreeBSD, which can be particularly useful for a package depending on a binary artifact.  Based on the [script](https://github.com/JuliaMath/SpecialFunctions.jl/blob/d8727a02d4cc5b34d2b3a1525c86869ef2a1104a/.cirrus.yml) for `SpecialFunctions.jl`

PR #89, opened 2019-02-04
Add Cirrus script to test the package on FreeBSD

giordano:
Replaces #88


ararslan:
I don't know that MKL is available for FreeBSD? If it is then that seems reasonable.

ararslan:
Oh yeah, you're right, looks like there's no Conda for FreeBSD. I guess I've never noticed because I avoid Python at all costs.

PR #96, opened 2019-05-08
Docs: add example

heitorPB:


docs/src/examples.md
```diff
@@ -0,0 +1,42 @@
...
+t0 = 0
+tmax = t0 + N * Ts
+
+# time coordinate
+t = t0:Ts:tmax
+
+signal = sin.(2π * 60 * t) + .5 * sin.(2π * 120 * t)
+
+# Fourier Transform of it
+F = fft(signal) |> fftshift
```
ararslan:
```suggestion
F = fftshift(fft(signal))
```

docs/src/examples.md
```diff
@@ -0,0 +1,42 @@
...
+freqs = fftfreq(length(t), fs) |> fftshift
```
ararslan:
```suggestion
freqs = fftshift(fftfreq(length(t), fs))
```

heitorPB:
I agree with you. But the basic structure should not change enough to break a simple example like this one.

ararslan:
Sorry for the multi-year delay on this one, @heitorPB, and thanks for sticking with it!

docs/src/examples.md
```diff
@@ -0,0 +1,42 @@
...
+using Plots
+using FFTW
+using DSP
+
+# Number of points
+N = 2^12 - 1
+# Sample spacing
+Ts = 1 / (1.1 * N)
+# Sample rate
+fs = 1 / Ts
```
palday, line 21: together, this is equivalent to saying that the sample rate is `1.1 * N` Hz, i.e. 4504.5

docs/src/examples.md
```diff
@@ -0,0 +1,42 @@
...
+tmax = t0 + N * Ts
+
+# time coordinate
+t = t0:Ts:tmax
+
+signal = sin.(2π * 60 * t) + .5 * sin.(2π * 120 * t)
+
+# Fourier Transform of it
+F = fftshift(fft(signal))
+freqs =  fftshift(fftfreq(length(t), fs))
```
ararslan, line 33: It could be worthwhile to add a comment explaining why `fftshift` is being used here.

docs/src/examples.md
```diff
@@ -0,0 +1,42 @@
...
+
+# Plot
+time_domain = plot(t, signal, title="Signal", xlims=(0, 4 / 60), xlabel="time (s)", label="")
+freq_domain = plot(freqs, abs.(F), title="Spectrum", xlims=(0, 200), xlabel="frequency (Hz)", label="")
+plot(time_domain, freq_domain, layout = 2)
+savefig("Wave.pdf")
```
ararslan, line 39: We could consider make this a Documenter `@example` block, which will render the output into the document.

ararslan (commented):
Oops, I started a review hours ago and forgot to push submit

docs/src/examples.md
```diff
@@ -0,0 +1,40 @@
...
+# Start time
+t0 = 0
+tmax = t0 + N * Ts
+
+# time coordinate
+t = t0:Ts:tmax
+
+signal = sin.(2π * 60 * t) + .5 * sin.(2π * 120 * t)
+
+# Fourier Transform of it
```
heitorPB, line 29:
```suggestion
# The `fft` function calculates the (discrete) Fourier transform of its input.
# The first half of the returned array contains the positive frequencies, while
# the second half contains the negative ones. For visualization purposes, we
# rearrange the array to have the zero-frequency at the center.

```

docs/src/examples.md
```diff
@@ -0,0 +1,42 @@
...
+# Sample rate
+fs = 1 / Ts
+# Start time
+t0 = 0
+tmax = t0 + N * Ts
+
+# time coordinate
+t = t0:Ts:tmax
+
+signal = sin.(2π * 60 * t) + .5 * sin.(2π * 120 * t)
```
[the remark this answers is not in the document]
heitorPB: I wrote this code ages ago and have been out of DSP for a while, but isn't the sampling rate `Ts` in this example and `60` the frequency?

The steps in the time coordinate are `Ts`, that means the sampling frequency, right?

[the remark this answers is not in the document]
ararslan: Sampling rate and frequency, at least as I learned the terms, are equivalent and refer to the number of samples per second, here called `fs`. `Ts` is as you said, the steps in the time coordinates, i.e. the time elapsed between successive samples, also called the sampling interval.

In this particular line of code, 60 is acting as the frequency in the sine wave formula (`amplitude * sin(2π * frequency * time + phase)`). But for the quantities used in the overall example to be consistent with one another, it should be `fs`, which isn't 60. If `fs` was 60 then `Ts = 1/60 ≈ 0.01667`, but `Ts` is defined above this line to be about 0.00022, which means that `fs` can't be 60.

docs/src/examples.md
ararslan: This comment is great, thank you!

docs/src/examples.md
ararslan: (Btw, you don't have to make this change as part of this PR unless you want to. I mention it more as a note.)

docs/src/examples.md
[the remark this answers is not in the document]
ararslan: Okay, I think I get it now: the actual wave in your example is 60 _cycles_ per second and it's sampled at 4504.5 _samples_ per second. Thanks, I appreciate the explanation and apologies for the confusion!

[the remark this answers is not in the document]
ararslan: My confusion stemmed from the (valid) use of Hertz for both cycles and samples; I only encounter the latter in my work so the former hadn't occurred to me.

Perhaps a comment could be worthwhile?
```suggestion
# The underlying signal here is the sum of a sine wave at 60 cycles per second
# and its second harmonic (120 cycles per second) at half amplitude. We have
# discrete observations (samples) of this signal at each time `t`, with `fs`
# samples per second.

signal = sin.(2π * 60 * t) + .5 * sin.(2π * 120 * t)
```

ararslan (approved):
Thanks so much for the contribution, @heitorPB!

PR #97, opened 2019-05-08
Add Project.toml and require Julia 1.0

giordano:


.travis.yml
```diff
@@ -20,3 +19,15 @@ notifications:
...
   - julia -e 'using Pkg; ps=Pkg.PackageSpec(name="Documenter", version="0.19"); Pkg.add(ps); Pkg.pin(ps); Pkg.dir("FFTW"); include(joinpath("docs", "make.jl"))'
+
+jobs:
+  include:
+    - stage: "Documentation"
+      julia: 1.0
+      os: linux
+      script:
+        - julia --project=docs/ -e 'using Pkg; Pkg.instantiate();
+                                               Pkg.develop(PackageSpec(path=pwd()));
```
ararslan: `develop` should come before `instantiate`, otherwise weird version resolution issues can happen

.travis.yml
```diff
@@ -20,3 +19,15 @@ notifications:
 after_success:
   - julia -e 'using Pkg; Pkg.add("Coverage"); Pkg.dir("FFTW"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
   - julia -e 'using Pkg; ps=Pkg.PackageSpec(name="Documenter", version="0.19"); Pkg.add(ps); Pkg.pin(ps); Pkg.dir("FFTW"); include(joinpath("docs", "make.jl"))'
```
ararslan: You can remove this if you're adding a documentation step

Project.toml
```diff
@@ -0,0 +1,23 @@
+name = "FFTW"
+uuid = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
+version = "0.3.0-DEV"
```
ararslan: The -DEV part isn't needed
```suggestion
version = "0.3.0"
```

Project.toml
```diff
@@ -0,0 +1,23 @@
...
+Conda = "8f4d0f93-b110-5947-807f-2305c1781a2d"
+Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+
+[compat]
+AbstractFFTs = "≥ 0.3.0"
+BinaryProvider = "≥ 0.3.0"
+julia = "≥ 1.0.0"
```
ararslan:
```suggestion
julia = "^1.0"
```

docs/Project.toml
```diff
@@ -0,0 +1,6 @@
+[deps]
+Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
```
ararslan:
```suggestion
```
This should only be in there if a manifest is committed that tells it that the path is relative to this directory, otherwise documentation will always try to use the release version of FFTW.

ararslan:
Looks like CI is complaining because we still have `using Compat` but Compat isn't in the Project.toml `[deps]`. Uses of Compat can probably just be removed now.

Project.toml
giordano: I was under the impression that only the release commit should have a non-dev version, apparently I misunderstood this.

.travis.yml
giordano: Guys, I can't follow all [changes to documentation](https://juliadocs.github.io/Documenter.jl/v0.20/man/hosting/#.travis.yml-Configuration-1) :sweat_smile:

Project.toml
ararslan: Nope. When checking something out to a particular branch (including master), Pkg just recognizes the version as `x.y.z+`, where `x.y.z` is what's declared in the Project.toml `version`. The only time `version` actually matters is for registering and for compatibility bounds.

.travis.yml
ararslan: That should probably be updated; I'm not sure why it's like that.

docs/Project.toml
```diff
@@ -0,0 +1,5 @@
+[deps]
+Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+
+[compat]
+Documenter = "~0.20"
```
ararslan:
```suggestion
Documenter = "~0.22"
```
Might as well go newest

Project.toml
giordano: I've used the `gen_project.jl` script from `Pkg.jl` to generate this, perhaps it could be improved?  (BTW, what's the difference?)

.travis.yml
```diff
@@ -4,7 +4,6 @@ os:
   - linux
   - osx
 julia:
-  - 0.7
   - 1.0
```
ararslan:
```suggestion
  - 1.0
  - 1.1
```
Since 1.1 was added to AppVeyor

.travis.yml
giordano: It's updated in the latest version (I linked docs for v0.20), but I copied from another Travis script of mines that was done some time ago.

PR #98, opened 2019-05-15
Create CITATION.bib

asinghvi17:
Creates a minimal `CITATION.bib`.  As of now I only have the 2005 paper; however, it would be easy to add the 1999, 1998 and 1997 ones.

### DIFF
--- a/CITATION.bib
+++ b/CITATION.bib
@@ -1,11 +1,12 @@
 % If you wish to cite FFTW in a general publication, we recommend this paper:
 @Article{FFTW.jl-2005,
-  author = 	 {Frigo, Matteo and Johnson, Steven~G.},
-  title = 	 {The Design and Implementation of {FFTW3}},
-  journal = 	 {Proceedings of the IEEE},
-  year = 	 2005,
-  volume =	 93,
-  number =	 2,
-  pages =	 {216--231},
-  note =	 {Special issue on ``Program Generation, Optimization, and Platform Adaptation''}
+  author   =  {Frigo, Matteo and Johnson, Steven~G.},
+  title    =  {The Design and Implementation of {FFTW3}},
+  journal  = 	{Proceedings of the IEEE},
+  year     = 	2005,
+  volume   =	93,
+  number   =  2,
+  pages    =  {216--231},
+  note     =	{Special issue on ``Program Generation, Optimization, and Platform Adaptation''},
+  doi      =  {10.1109/JPROC.2004.840301}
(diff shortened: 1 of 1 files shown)


giordano:
It has a DOI that could be added with the `doi` field: `10.1109/JPROC.2004.840301`

asinghvi17:
@giordano done

giordano:
Are you using tabs?  It looked better before the reformat (remember the `[skip ci]` :wink:)

asinghvi17:
Should look better now, some tabs must have snuck past :)

(Also, does `[skip ci]` have to be in the commit message?  I put it in the description, but it seems to have triggered CI anyway)

stevengj:
What is CITATION.bib for?

The relevant standard here seems to be a `codemeta.json` file (https://codemeta.github.io/), no?  And FFTW already has this: https://github.com/FFTW/fftw3/blob/master/codemeta.json … not sure why FFTW.jl needs one too?

asinghvi17:
There was a discussion over on the `academia` channel of the JuliaLang slack, wherein we'd agreed on a `CITATION.bib` file format and a function that could look at your `Project.toml` and automatically compile a `.bib` file with citations for all the packages you'd used.

It's just another standard, I guess - though we have `JuliaDiffEq` and `JuliaDynamics` (and many others) on board.  

> The relevant standard here seems to be a codemeta.json file 

I wasn't aware of that!  That's certainly something to look into - maybe such a function could support both formats :)

asinghvi17:
Also, to quote Viral:

> The immediate hope is to encourage and make it easy to cite. I have mentioned this on https://julialang.org/research. As these start becoming common, the citation information will make its way into pkg.julialang.org, package docs, and other places.

stevengj:
It's frustrating to have a Julia-specific standard here, but since it is documented I guess we can add it.

@ViralBShah, can you comment on why you didn't adopt codemeta?

ViralBShah:
I think this started out as a stopgap and people have been discussing many alternatives but haven't reached a conclusion. We could still decide in favour of something else.

ViralBShah:
Cc @stefankarpinski

stevengj:
Codemeta seems pretty widely adopted. Is there any competing standard, or reason not to adopt the same thing?

stevengj:
I opened up a discussion on discourse — https://discourse.julialang.org/t/replacing-citation-bib-with-a-standard-metadata-format

Please comment there.

PR #105, opened 2019-07-30
partr thread support

### LINKED ISSUE #66: FFTW planner is not thread safe.
Moved from https://github.com/JuliaLang/julia/issues/17972

---------------------

Having fft/bfft/ifft/etc. calls in a threaded region exposes the fact that the planning portion of the fft call is not thread-safe in fftw.  Below is a potential patch to dft.jl if you just want to lock the planner.

```
global planner_mutex = nothing

function __init__()
    global planner_mutex = Base.Threads.Mutex()
end

# implementations only need to provide plan_X(x, region)
# for X in (:fft, :bfft, ...):
for f in (:fft, :bfft, :ifft, :fft!, :bfft!, :ifft!, :rfft)
    pf = Symbol("plan_", f)
    @eval begin
        function $f(x::AbstractArray)
            Base.Threads.lock!(planner_mutex)
            the_plan = $pf(x)
            Base.Threads.unlock!(planner_mutex)
            the_plan * x
        end
        function $f(x::AbstractArray, region)
            Base.Threads.lock!(planner_mutex)
            the_plan = $pf(x, region)
            Base.Threads.unlock!(planner_mutex)
            the_plan * x
        end
        function $pf(x::AbstractArray; kws...)
            Base.Threads.lock!(planner_mutex)
            the_plan = $pf(x, 1:ndims(x); kws...)
            Base.Threads.unlock!(planner_mutex)
            the_plan
        end
    end
end
```

stevengj:
Support partr threads (JuliaLang/julia#31398) on the latest Julia master branch via FFTW/fftw3#175.

When it is using the partr backend, by default it sets the number of FFTW "threads" to 4*nthreads — we want to spawn more tasks than we have threads to help with load balancing if other stuff is running.

It also enables the [thread-safe FFTW planner](http://www.fftw.org/fftw3_doc/Thread-safety.html) (which puts a mutex lock around plan creation).  In the longer run, it would be better to do the locking on the Julia side, since presumably using a Julia lock would allow other Julia tasks to wake up. Closes #66.

To get the full benefit of threading, you should precompute the FFT plan via `p = plan_fft(array)` [or `p = plan_fft(array, flags=FFTW.MEASURE)` or `p = plan_fft(array, flags=FFTW.PATIENT)` if you want it to self-optimize the plan], rather than calling `fft(array)` directly.

cc @vtjnash, who tested this at JuliaCon.  I think I included all of the fixes we did on your machine.


stevengj:
Hmm, AppVeyor is crashing on x64 with:
```
Please submit a bug report with steps to reproduce this fault, and any error messages that follow (in their entirety). Thanks.
Exception: EXCEPTION_ACCESS_VIOLATION at 0x1ed37c3b -- VMOVAPD_LD at C:\projects\fftw-jl\deps\usr\bin\libfftw3-3.dll (unknown line)
in expression starting at C:\projects\fftw-jl\test\runtests.jl:10
VMOVAPD_LD at C:\projects\fftw-jl\deps\usr\bin\libfftw3-3.dll (unknown line)
hc2cfdftv_20 at C:\projects\fftw-jl\deps\usr\bin\libfftw3-3.dll (unknown line)
apply_extra_iter at C:\projects\fftw-jl\deps\usr\bin\libfftw3-3.dll (unknown line)
apply_dit_dft at C:\projects\fftw-jl\deps\usr\bin\libfftw3-3.dll (unknown line)
.text at C:\projects\fftw-jl\deps\usr\bin\libfftw3-3.dll (unknown line)
unsafe_execute! at C:\projects\fftw-jl\src\fft.jl:407 [inlined]
* at C:\projects\fftw-jl\src\fft.jl:729
```
The AppVeyor tests are single threads (`JULIA_NUM_THREADS` should be `1`, the default), so it is not using any of the new partr code in this PR.

I can only speculate that something went wrong in the x64 BinaryBuilder cross-compile with the latest build tools…

ViralBShah:
Is fixing #66 relevant for this PR?

stevengj:
@ViralBShah, this PR fixes #66, which seems like a good idea for using FFTW in an environment with more pervasive threading but is not strictly required for partr support.

stevengj:
Switched to the new build from JuliaPackaging/Yggdrasil#53, which uses different compiler versions, thanks to @staticfloat.  Let's see if that helps.

staticfloat:
Sorry, I had a typo in the `build.jl` I gave you.  Fixed now.

PR #119, opened 2019-10-07
Use parametric type for `region` field

jipolanco:
I was wondering if there's any reason for the `region` field of the FFTWPlan types to be `Any`.

This PR sets `region` to a concrete (parametric) type. Tests pass, at least on my machine with Julia 1.3.0-rc3.

Feel free to ignore this if there's something I'm missing.

### DIFF
--- a/src/fft.jl
+++ b/src/fft.jl
@@ -226,7 +226,7 @@ end
 abstract type FFTWPlan{T<:fftwNumber,K,inplace} <: Plan{T} end
 for P in (:cFFTWPlan, :rFFTWPlan, :r2rFFTWPlan) # complex, r2c/c2r, and r2r
     @eval begin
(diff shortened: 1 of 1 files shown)


mcabbott:
Ideally this would extend up to the types in https://github.com/JuliaMath/AbstractFFTs.jl, `abstract type Plan{T,R} end`.

stevengj:
It's not clear that the abstract `Plan` needs this, since there are no abstract `Plan` methods that use the region info.

stevengj:
LGTM.

mcabbott:
My reason to want it in the AbstractFFTs type is that it would enable other packages, which depend only on that one, to rely on this information. (My original example involved NamedDims.jl.)

PR #122, opened 2019-10-25
Upgrade to Julia 1.3+, use JLL package to provide FFTW

staticfloat:
This is the first package where I've tried to mix JLL packages with more
custom installation procedures, but it seems to more-or-less work.


staticfloat:
Appveyor 1.3 x86 failure will be fixed in rc5

ararslan:
Lookin' good! Would it be possible to use a BinaryBuilder-provided MKL upon request instead of using Conda?

staticfloat:
> Lookin' good! Would it be possible to use a BinaryBuilder-provided MKL upon request instead of using Conda?

https://github.com/JuliaBinaryWrappers/MKL_jll.jl/releases/tag/MKL-v2019.0.117%2B0

ararslan:
Yeah, I know it exists, I'm just not sure how one would restructure the provider choice since it's actually loading a package rather than just picking up an existing library.

staticfloat:
I think just like how we do this:

https://github.com/JuliaMath/FFTW.jl/pull/122/files#diff-ed74277d7f2e98f12cf51e6b8be90705R44-R48

We could generate a few lines to load in `MKL_jll`.

Issue #118, opened 2019-10-02
Don't use Conda


Keno:
Installing FFTW.jl installs Conda, apparently to be able to test MKL on CI. This should be unnecessary. We have a MKL in Yggradsil, so we should be able to just install it: https://github.com/JuliaPackaging/Yggdrasil/blob/master/M/MKL/build_tarballs.jl.

ChrisRackauckas:
If it's just for testing, why is it not a just a test dependency?

stevengj:
It's not just a test dependency — if you configure FFTW.jl with `JULIA_FFTW_PROVIDER=MKL`, and you don't have an MKL build of Julia, then it installs MKL for you with Conda.

Now that it is on Yggdrasil, I agree that we can just install it from there.

Once the whole "jll" mechanism lands, however, will this mean that we have to add a dependency on both FFTW.jll and MKL.jll, so that everyone has to install both binaries even if they are only using one?  I'm still pining for optional dependencies in Pkg.

ianshmean:
Some progress over in #127 but looks like we need `IntelOpenMP_jll` too

ianshmean:
#127 has been merged, so master no longer uses Conda. Ready for a release?

PR #127, opened 2019-12-02
use MKL_jll instead of Conda during build

ianshmean:
Trying to provide MKL through MKL_jll rather than Conda during build (#118)

Currently hits the following in Julia 1.3.0
```julia
julia> ENV["JULIA_FFTW_PROVIDER"]="MKL"
"MKL"

(FFTW) pkg> build
  Building FFTW → `~/Documents/GitHub/FFTW.jl/deps/build.log`
 Resolving package versions...

(FFTW) pkg> test
   Testing FFTW
 Resolving package versions...
Intel MKL FATAL ERROR: Cannot load libmkl_intel_thread.dylib.
ERROR: Package FFTW errored during testing
```


ianshmean:
Seems like we also need IntelOpenMP, and the IntelOpenMP [`build_tarball.jl`](https://github.com/JuliaPackaging/Yggdrasil/blob/master/I/IntelOpenMP/build_tarballs.jl) needs updating, and looks non-standard

giordano:
Pull request for IntelOpenMP at https://github.com/JuliaPackaging/Yggdrasil/pull/259

ianshmean:
All platforms that were being tested on master are passing.

However, Windows MKL wasn't being tested, but is failing with:
```
ERROR: LoadError: ArgumentError: FFTW plan applied to array with wrong memory alignment
```

I've [PR'd to add windows MKL tests to master](https://github.com/JuliaMath/FFTW.jl/pull/128), and it's still failing, but with a different error:
```
ERROR: LoadError: LoadError: could not load library "C:\Users\travis\.julia\conda\3\Library\bin\libmkl_rt"
The specified module could not be found.
```

ararslan:
Thanks for this. Seems the README is now outdated, as it says that it uses Conda. Would you be willing to update that as well, @ianshmean?

PR #138, opened 2020-02-21
nthreads check should be outside of static if

stevengj:
This line looks suspicious to me — seems like this needs to be a runtime check, not in `@static if`.

### DIFF
--- a/src/FFTW.jl
+++ b/src/FFTW.jl
@@ -50,13 +50,15 @@ function __init__()
     if stat == 0 || statf == 0
         error("could not initialize FFTW threads")
     end
-    @static if fftw_vendor == :fftw && nthreads() > 1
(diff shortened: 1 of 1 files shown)


ararslan:
Can `nthreads` be configured at runtime? I thought it was only set at startup.

stevengj:
It's only at startup, but doesn't `@static` mean that this check executes at lowering time (which happens when the module is precompiled), not at module-loading time?

ararslan:
Ah yeah, you're right.

jtravs:
Any chance we could get a release tagged with this fix in it? I was just bitten by this and wasted quite a lot of time.

ararslan:
I believe I was as well and didn't even realize it. #148

PR #147, opened 2020-04-23
Some CI improvements

giordano:
* Add CompatHelper
* Reduce frequency of TagBot to once per day
* Remove Cirrus script, it's not working properly lately
* Add FreeBSD to Travis
* Run Travis only on master and a few special-named branches
* Cache artifacts in Travis and AppVeyor


giordano:
Is there an easy way to exclude FreeBSD with MKL?

ararslan:
> Is there an easy way to exclude FreeBSD with MKL?

Should be able to use `exclude:` in `jobs:`, I believe.

giordano:
> Should be able to use exclude: in jobs:, I believe.

Does it work with environment variables?  I didn't try.  I went with the opposite way: `include`-ing the FreeBSD jobs :sweat_smile:

ararslan:
Yeah I believe it does.

giordano:
Looks like it doesn't?

giordano:
Shall I go back to `include`-ing or maybe I did something wrong now?

.travis.yml
```diff
@@ -21,7 +21,22 @@ notifications:
...
+
+cache:
+  directories:
+    - $HOME/.julia/artifacts
+
 jobs:
+  exclude:
+    - os: freebsd
+      env:
+        - JULIA_FFTW_PROVIDER=MKL
```
ararslan, line 39:
```suggestion
      env: JULIA_FFTW_PROVIDER=MKL
```
? Not sure it will make a difference but worth a try I guess

giordano: :open_mouth:

ararslan:
Looks like that worked! :tada:

PR #149, opened 2020-05-24
Add compatibility with `MKL_jll` v2020

giordano:


stevengj:
Would be nice if MKL_jll adopted semver, even if upstream does not, by mapping MKL upstream version numbers to sensible semver values.

PR #150, opened 2020-05-24
Add get_num_threads function

### LINKED ISSUE #117: `get` version of `FFTW.set_num_threads()`?
I want to temporarily change `FFTW.set_num_threads()` and restore what it was, but can't find `FFTW.get_num_threads()` or an equivalent to read before setting.

Does such a function exist?

galenlynch:
While it is possible to change the number of threads used by FFTW plans with
`set_num_threads`, it is impossible to see what the current value is. This is
made difficult because FFTW3 does not, as far as I know, provide a way to do
this. However, caching the last argument passed to `set_num_threads` as a
constant in FFTW.jl allows later access. I have added a constant,
`PLAN_NUM_THREADS` to keep track of the last argument to `set_num_threads`,
modified `set_num_threads` to update this constant, and added a function
`get_num_threads` to access the constant.

Closes #117

### DIFF
--- a/src/fft.jl
+++ b/src/fft.jl
@@ -108,6 +108,11 @@ const fftwSingle = Union{Float32,Complex{Float32}}
 const fftwTypeDouble = Union{Type{Float64},Type{Complex{Float64}}}
 const fftwTypeSingle = Union{Type{Float32},Type{Complex{Float32}}}
 
+# cached argument to set_num_threads
+
+# Default of 1 because set_num_threads is only called if Threads.nthreads() > 1
+const PLAN_NUM_THREADS = Ref(Int32(1))
+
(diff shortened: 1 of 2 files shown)


stevengj:
This is reasonable, but I would ideally like to fix this in the FFTW C library; otherwise in principle this could return the wrong result if a C library links to FFTW and changes this.

https://github.com/FFTW/fftw3/blob/516116615362334c2051a16ec1658a888573ab63/threads/api.c#L70-L81 is where a new function `int X(planner_nthreads)(void) { return X(the_planner)()->nthr; }` would go.

galenlynch:
I will attempt to add this to the FFTW C library instead of caching it in Julia (FFTW/fftw3#205).

PR #151, opened 2020-05-25
don't set nthreads > 1 by default

stevengj:
Defaulting to multi-threaded FFTW was causing too many slowdowns in code that was calling FFTW in parallel from multiple threads (where the right thing is to use the serial FFTW, especially given Julia's current thread-spawning overhead).

Closes #121.  Closes JuliaDSP/DSP.jl#339.

PR #154, opened 2020-06-01
Bump patch version

ararslan:
Patch bump seemed okay to me but a minor could also be in order if #151 has the potential to adversely affect any code that was assuming FFTW defaulted to multiple threads.


ararslan:
Actually, disregard this. The patch version was already bumped in #149.

Issue #141, opened 2020-03-10
Deadlock with multiple FFTW plan construction / destruction in serial and threaded runs


mfherbst:
Running the following code example causes a deadlock conjunction with Julia 1.3.1, DFTK 0.0.5 and FFTW 1.2.0. [Stacktrace](https://gist.github.com/mfherbst/4015ff891001597fccf66648abff1aa1) (after a Ctrl-C) indicates that the deadlock occurs in fftw. Notice, such behaviour is encountered both in serial and threaded runs.

Code example:
```
using DFTK    
    
function main()    
    for Ecut in 0.0 .+ collect(3:20)    
        println("Ecut = $Ecut")    
        a = 10.263141334305942    
        lattice = a / 2 .* [[0 1 1.]; [1 0 1.]; [1 1 0.]]    
        model = Model(lattice; terms=[Kinetic()], n_electrons=4)    
        basis = PlaneWaveBasis(model, Ecut)    
    end    
end    
    
main()   
```

Output / Stacktrace: https://gist.github.com/mfherbst/4015ff891001597fccf66648abff1aa1

Sorry about the complicated reproducer btw ... I tried to reduce as much as I could.

mfherbst:
Note: The issue also present on master as of now.

mfherbst:
Could be related to #135.

mfherbst:
The issue disappears when setting explicitly `FFTW.set_num_threads(1)`.

stevengj:
@vtjnash, any ideas?   FFTW has an internal pthreads mutex lock around plan creation and destruction, but I don't see how that could deadlock Julia…

EthanAnderes:
I think I'm seeing the same issue. Here is a reduced toy case.```
using FFTW 
import Base: * 

struct rFFT{nᵢ} end

@generated function plan(::Type{F}) where {nᵢ, F<:rFFT{nᵢ}}
    X  = zeros(Float64, nᵢ...) 
    FT = plan_rfft(X; flags=FFTW.MEASURE, timelimit=10)
    return :( $FT )
end

nᵢ = (256, 50)
FT = rFFT{nᵢ}
matx = rand(Float64, nᵢ...)

plan(FT) * matx
```

The last line hangs with no CPU activity coming from Julia. If, however, I replace the 
last line with the following two lines 

```
*(::Type{FT}, x::Array) where FT<:rFFT = plan(FT) * x
FT * matx
```

then I get this segfault. 

```
[...]

PR #157, opened 2020-06-21
Trying a delayed finalizer approach to fix threaded deadlock (#141)

ianshmean:
Edit: Seems the julia crash was poor const handling that's now fixed.

Trying out @vtjnash’s suggestion from #141

> For now, I recommend only doing trivial things in a finalizer, like appending it to a queue to be cleaned up later.

This PR "fixes" the deadlock seen in the main DFTK example in #141 in the sense that the example finishes, ~~but when the queued & delayed finalizer runs, Julia crashes.~~ It's ~~obviously~~ _probably_ not a proper fix yet as I'm sure there's a better/correct/functional way to do this, but I thought I'd share my workings

```
using DFTK    
DFTK.FFTW.set_num_threads(Threads.nthreads() * 4)
    
function main()    
    for Ecut in 0.0 .+ collect(3:20)    
        println("Ecut = $Ecut")    
        a = 10.263141334305942    
        lattice = a / 2 .* [[0 1 1.]; [1 0 1.]; [1 1 0.]]    
        model = Model(lattice; terms=[Kinetic()], n_electrons=4)    
        basis = PlaneWaveBasis(model, Ecut)    
    end    
end    
    
@show Threads.nthreads()
main() 
```
```
Threads.nthreads() = 6
Ecut = 4.0
Ecut = 5.0
Ecut = 6.0
Ecut = 7.0
Ecut = 8.0
Ecut = 9.0
Ecut = 10.0
Ecut = 11.0
Ecut = 12.0
Ecut = 13.0
Ecut = 14.0
Ecut = 15.0
Ecut = 16.0
Ecut = 17.0
Ecut = 18.0
Ecut = 19.0
Ecut = 20.0
```

cc. @stevengj @vtjnash @mfherbst

### DIFF
--- a/src/fft.jl
+++ b/src/fft.jl
@@ -255,12 +255,26 @@ for P in (:cFFTWPlan, :rFFTWPlan, :r2rFFTWPlan) # complex, r2c/c2r, and r2r
 end
 
 const DESTROY_QUEUE = Vector{FFTWPlan}(undef, 0)
-const DESTROY_QUEUE_TIMER = Timer((t) -> destroy_plan.(DESTROY_QUEUE), 0)
+
+function destroy_queued_plans(t)
+    while length(DESTROY_QUEUE) > 0
+        p = popfirst!(DESTROY_QUEUE)
+        destroy_plan(p)
+    end
+    return nothing
+end
+
+const DESTROY_QUEUE_TIMER = Ref{Timer}(Timer(destroy_queued_plans, 0))
 
 function queue_destroy_plan(p::FFTWPlan)
     push!(DESTROY_QUEUE, p) # add plan to destroy queue
-    DESTROY_QUEUE_TIMER = Timer((t) -> destroy_plan.(DESTROY_QUEUE), 5) # delay destroying plans by another 5 seconds
+    close(DESTROY_QUEUE_TIMER[])
+    DESTROY_QUEUE_TIMER[] = Timer(destroy_queued_plans, 5) # delay destroying plans by another 5 seconds
 end
+
+
+    
+
 
 size(p::FFTWPlan) = p.sz
 
--- a/src/fft.jl
+++ b/src/fft.jl
@@ -258,8 +258,7 @@ const DESTROY_QUEUE = Vector{FFTWPlan}(undef, 0)
 
 function destroy_queued_plans(t)
     while length(DESTROY_QUEUE) > 0
-        p = popfirst!(DESTROY_QUEUE)
-        destroy_plan(p)
+        destroy_plan(popfirst!(DESTROY_QUEUE)) # start with oldest
     end
     return nothing
 end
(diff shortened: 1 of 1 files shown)


ianshmean:
If this queue & delayed freeing approach is valid, I think the key is figuring out the right thing to trigger `destroy_queued_plans` on. 

If this timer-based approach isn't appropriate, is there another point that would be reliably logical to call `destroy_queued_plans`?

ianshmean:
bump. 
It'd be good to figure out if this is acceptable

@giordano did you manage to test it on your deadlocking code?

giordano:
> @giordano did you manage to test it on your deadlocking code?

Yes, I confirm that, in the code where I was experiencing a deadlock, with this PR and `FFTW.set_num_threads(8)`, now `fft` doesn't freeze and instead the computation goes on!

ianshmean:
@stevengj @vtjnash is this an acceptable solution?

src/fft.jl
```diff
@@ -254,6 +254,23 @@ for P in (:cFFTWPlan, :rFFTWPlan, :r2rFFTWPlan) # complex, r2c/c2r, and r2r
...
+const DESTROY_QUEUE = Vector{FFTWPlan}(undef, 0)
+
+function destroy_queued_plans(t)
+    while length(DESTROY_QUEUE) > 0
+        destroy_plan(popfirst!(DESTROY_QUEUE)) # start with oldest
+    end
+    return nothing
+end
+
+const DESTROY_QUEUE_TIMER = Ref{Timer}(Timer(destroy_queued_plans, 0))
```
stevengj, line 266: This doesn't really seem reasonable to me; if there was a deadlock before, then this just introduces a race condition, no?  Also, a `Timer` with an interval of 0 seems like a spinloop, which troubles me.

ianshmean: That timer has a delay of 0, so calls once  immediately. Interval is a kwarg that’s not used here.

My take on the original suggestion is to delay the destroying of plans until general fft planning has ended. But I don’t think the timer approach is right. Is there another event that could trigger destroying the queue, other than the finalizer?

vtjnash: You could try to drain the queue manually after any operation returns. Since the timer will wait for `yield`, I think you can usually use a zero-length timeout (basically just an idle event, but we don't wrap those in libuv)

ianshmean: I tried changing the 5 second timer to 0 seconds and the deadlock with the MWE returns.
1 second gives the same as 5 seconds, but 0.1 seconds results in deadlock. It seems too unstable to be considered a fix

vtjnash:
Ah, true. The issue is that it’s using a non-reentrant lock, but that means we have difficulty since we’re using Julia’s task system, but not our locks. (So also might run into similar deadlocks if two threads try to FFT at the same time, possibly?)

ianshmean:
Is there anything on the horizon that could be/support a viable fix for this? Feels like the issue itself is deadlocked..

Octogonapus:
Have there been any updates on this? How can we make progress?

ianshmean:
@vtjnash I forget if we came up with any next steps when we discussed this at juliacon?

vtjnash:
I think we said a couple things (or at least, I thought them and meant to say them) about what's going on here under the hood. I looked into this a bit closer today, and realized that we missed that we can call `set_planner_hooks` when adding the call to `fftw_threads_set_callback`. Per the warning in the documentation (http://www.fftw.org/fftw3_doc/Thread-safety.html#Thread-safety), this can cause "the worst of all worlds". What may be desirable here is to tell fftw to use our `Base.ReentrantLock` instead in these callback hooks. When we unlock this lock, we can then flush the list of pending finalizers (instead of using the timer here).

(In the less immediate future, we should run all finalizers on a separate thread, which would have avoided this from being a problem in the first place)

stevengj:
#160 uses `ReentrantLock` and gets rid of the FFTW locks, so it should fix this issue.

PR #160, opened 2020-08-17
use Julia ReentrantLock rather than fftw_make_planner_threadsafe

### LINKED ISSUE #141: Deadlock with multiple FFTW plan construction / destruction in serial and threaded runs

stevengj:
Should fix #141.  Closes #157.

### DIFF
--- a/src/FFTW.jl
+++ b/src/FFTW.jl
@@ -52,14 +52,40 @@ function __init__()
     end
     @static if fftw_vendor == :fftw
         if nthreads() > 1
-            ccall((:fftw_make_planner_thread_safe,  libfftw3),  Cvoid, ())
-            ccall((:fftwf_make_planner_thread_safe, libfftw3f), Cvoid, ())
             cspawnloop = @cfunction(spawnloop, Cvoid, (Ptr{Cvoid}, Ptr{Cvoid}, Csize_t, Cint, Ptr{Cvoid}))
             ccall((:fftw_threads_set_callback,  libfftw3),  Cvoid, (Ptr{Cvoid}, Ptr{Cvoid}), cspawnloop, C_NULL)
             ccall((:fftwf_threads_set_callback, libfftw3f), Cvoid, (Ptr{Cvoid}, Ptr{Cvoid}), cspawnloop, C_NULL)
         end
     end
 end
+
+# most FFTW calls other than fftw_execute should be protected by a lock to be thread-safe
+const fftwlock = ReentrantLock()
+
+# protect thread-safety of expressions or whole functions by fftwlock:
+macro exclusive(ex)
+    if Meta.isexpr(ex, :function) || (Meta.isexpr(ex, :(=)) && Meta.isexpr(ex.args[1], :call))
+        newbody = quote
+            try
+                lock(fftwlock)
+                $(esc(ex.args[2]))
+            finally
+                unlock(fftwlock)
+            end
+        end
+        Expr(:function, esc(ex.args[1]), newbody)
+    else
+        return quote
+            try
+                lock(fftwlock)
+                $(esc(ex))
+            finally
+                unlock(fftwlock)
+            end
+        end
+    end
+end
+
 
 include("fft.jl")
 include("dct.jl")
--- a/src/FFTW.jl
+++ b/src/FFTW.jl
@@ -66,8 +66,8 @@ const fftwlock = ReentrantLock()
 macro exclusive(ex)
     if Meta.isexpr(ex, :function) || (Meta.isexpr(ex, :(=)) && Meta.isexpr(ex.args[1], :call))
         newbody = quote
+            lock(fftwlock)
             try
-                lock(fftwlock)
                 $(esc(ex.args[2]))
             finally
                 unlock(fftwlock)
@@ -77,8 +77,8 @@ macro exclusive(ex)
         Expr(:function, esc(ex.args[1]), newbody)
     else
         return quote
+            lock(fftwlock)
             try
-                lock(fftwlock)
                 $(esc(ex))
             finally
                 unlock(fftwlock)
--- a/src/FFTW.jl
+++ b/src/FFTW.jl
@@ -71,6 +71,7 @@ macro exclusive(ex)
                 $(esc(ex.args[2]))
             finally
                 unlock(fftwlock)
+                destroy_deferred() # deallocate plans queued to be destroyed while we held the lock
             end
         end
         Expr(:function, esc(ex.args[1]), newbody)
@@ -85,7 +86,6 @@ macro exclusive(ex)
         end
     end
 end
-
 
 include("fft.jl")
 include("dct.jl")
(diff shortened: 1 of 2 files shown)


vtjnash:
This is a large portion of it, definitely! Though doesn't yet fix the deadlock. While this is a re-entrant lock, we shouldn't ever see it be re-entered from a given thread (whether on the same or a different task) as fftw doesn't like that. However, we might see that some `finalizer` tries to acquire this lock. To protect against that, the finalizers also need to be delayed until the unlock call here, to ensure that they destroy call doesn't get blocked.

stevengj:
I don't quite follow you.  If a given thread holds the lock and FFTW is busy calling some thread-unsafe routine (e.g. creating or destroying a thread), how could a finalizer concurrently execute `fftw_destroy_plan` on the same thread?

src/FFTW.jl
```diff
@@ -52,15 +52,41 @@ function __init__()
...
+# most FFTW calls other than fftw_execute should be protected by a lock to be thread-safe
+const fftwlock = ReentrantLock()
+
+# protect thread-safety of expressions or whole functions by fftwlock:
+macro exclusive(ex)
+    if Meta.isexpr(ex, :function) || (Meta.isexpr(ex, :(=)) && Meta.isexpr(ex.args[1], :call))
+        newbody = quote
+            try
+                lock(fftwlock)
+                $(esc(ex.args[2]))
```
vtjnash, line 71: For the sake of demonstration purposes only, consider all of the places we'll call `fftwf_destroy_plan` from, such as right here:
```suggestion
                GC.gc()
                $(esc(ex.args[2]))
```

For stress-testing, you actually want this in `spawnloop` though, since it's in that location that causes fftw to become deadlocked.

stevengj:
I see, so the scenario we are worried about is something like:

1. User starts creating a plan in a thread.  Acquires lock.
2. FFTW planner, while exploring different algorithms, calls `spawnloop`.
3. During Julia `spawnloop` execution, one of the spawned threads runs GC and tries to destroy an old plan.  Deadlocks.

stevengj:
Maybe when `fftwlock` is acquired, it should set a global flag that will cause all `destroy_plan` calls to be deferred to a queue, and that it can flush the queue right before it releases the lock?

That will get rid of all the business with a timer in #157, and allow garbage collection to function normally most of the time.

vtjnash:
That sounds great

stevengj:
Okay, I updated the thread to defer plan destruction while the lock is held.

(The plan-destruction queue itself requires a lock, and the logic of flushing the queue is a bit tricky to avoid deadlocks or race conditions, so I'd appreciate another pair of eyes to double-check my logic.)

src/fft.jl
```diff
@@ -258,18 +257,70 @@ size(p::FFTWPlan) = p.sz
...
+        end
+    finally
+        unlock(deferred_destroy_lock)
+    end
+end
+
+@exclusive destroy_plan(plan::FFTWPlan) = unsafe_destroy_plan(plan)
+
+function maybe_destroy_plan(plan::FFTWPlan)
+    if trylock(fftwlock)
```
vtjnash, line 300: This is re-entrant, so I think this’ll succeed, while the internal lock inside destroy will hang

src/fft.jl
```diff
@@ -258,18 +257,70 @@ size(p::FFTWPlan) = p.sz
...
+        try
+            unsafe_destroy_plan(plan)
+        finally
+            unlock(fftwlock)
+        end
+    else
+        try
+            lock(deferred_destroy_lock)
```
vtjnash, line 308: This lock needs to acquired first, so that another thread can’t release fftwlock and then skip past destroy_deferred before we get around to acquiring this lock

src/FFTW.jl
```diff
@@ -52,15 +52,42 @@ function __init__()
...
+            finally
+                unlock(fftwlock)
+                destroy_deferred() # deallocate plans queued to be destroyed while we held the lock
```
vtjnash, line 74: Perhaps slightly cheaper to move this before the unlock, since we may try immediately to reacquire it?

src/fft.jl
stevengj: Can I just use a non-reentrant lock?

src/FFTW.jl
stevengj: I had that originally, but that creates a possible memory leak — we destroy_deferred, then another finalizer runs and defers destroy_plan, then we unlock, leaving a non-empty list of plans to destroy.

src/fft.jl
stevengj: Right.

src/fft.jl
stevengj: Kind of annoying that Julia doesn't provide a non-reentrant lock other than `SpinLock`.   I suppose I could set a flag of some sort in addition to the lock…

stevengj: (There is no internal lock inside `unsafe_destroy_plan` now that I've turned off `fftw_make_planner_threadsafe`; the danger is that it will mutate FFTW internal state in a way that will corrupt the running planner.)

stevengj: I pushed an update with a `trylock_nonreentrant` function that should fix this?

src/FFTW.jl
```diff
@@ -52,15 +52,42 @@ function __init__()
...
 
+# most FFTW calls other than fftw_execute should be protected by a lock to be thread-safe
+const fftwlock = ReentrantLock()
+
+# protect thread-safety of expressions or whole functions by fftwlock:
+macro exclusive(ex)
+    if Meta.isexpr(ex, :function) || (Meta.isexpr(ex, :(=)) && Meta.isexpr(ex.args[1], :call))
+        newbody = quote
+            try
+                lock(fftwlock)
```
vtjnash, line 70: I think it's more common to switch the order of these try/lock calls (https://stackoverflow.com/a/10868476), though it shouldn't make a functional difference.

stevengj: Will do

stevengj:
Now that I think of it, probably it is actually safe to use the regular `trylock` in `maybe_destroy_plan` — the FFTW planner is not thread-safe, but it *is* re-entrant (and in fact we call it recursively inside FFTW), and the same should be true for `destroy_plan`.

So, it should be fine to destroy a plan in the thread that acquired the planner lock, even if that occurs during planning.

PR #162, opened 2020-08-20
uses a spintest to avoid task switches in finalizers

### LINKED ISSUE #161: Concurrency bugs in the latest release
With the release of v1.2.3, I am now seeing a lot of these errors:
```
error in running finalizer: ErrorException("task switch not allowed from inside gc finalizer")
error in running finalizer: ErrorException("val already in a list")
```

We use FFTW on multiple threads. Is this a supported use case?

I would have liked to attach a small example to reproduce the problem, but I can't figure one out. Maybe it would help if others commented with ideas on how these errors could happen.

stevengj:
Hopefully fixes #161


Octogonapus:
This fixes #161 for my use case.

PR #171, opened 2020-12-20
Add get_num_threads

### LINKED ISSUE #117: `get` version of `FFTW.set_num_threads()`?

galenlynch:
This commit adds `get_num_threads`, which returns the number of threads used by the planner, and is the complement to `set_num_threads`. This simply wraps the function `fftw_planner_nthreads`, which was [newly added to fftw in version 3.3.9](https://github.com/FFTW/fftw3/blob/34082eb5d6ed7dc9436915df69f376c06fc39762/NEWS#L3).

This PR will not work until JuliaPackaging/Yggdrasil#2291 is merged, and the resulting artifact is used for testing.

Closes #117


galenlynch:
This is now working locally for me, and tests are passing, when used with FFTW_jll v3.3.9+7.

stevengj:
Can you bump the version requirement on FFTW_jl in Project.toml?

galenlynch:
Yeah I wasn't sure what to do about that... I don't think you can specify a build number in Project.tomls. If I just change the `compat` section so that FFTW_jll's version is "3.3.9+7" I get the following error when trying to use FFTW:

```
ERROR: Could not parse compatibility version for dependency `FFTW_jll`
```

As long as people update their packages this _should_ work since they will get the most recent build of `FFTW_jll`, but I don't know how to make sure they get v3.3.9+7 (or later).

I can update the version requirement for FFTW_jll to be `3.3.9`, but that won't always work if people are using an older FFTW_jll build. Would that suffice?

giordano:
> I don't think you can specify a build number in Project.toml.

That's another reason why we want to change the versioning system of JLL package (and the reason why I was waiting to merge the PR)

galenlynch:
I imagine that not being able to specify the build number will not be a problem for people, unless they have pinned or dev'd `FFTW_jll`.

In order to make sure that it works, it seems like you would need to create a new FFTW_jll version v3.3.10 that corresponds to fftw3 v3.3.9, because FFTW_jll v.3.3.9 has been wrapping a pre-release version of fftw v3.3.9 for a while (likely for a good reason), meaning that there is currently no way to specify the released fftw 3.3.9 artifact with FFTW_jll's version.

I'm still not clear on how package maintainers will be able to specify which artifact version they want if the _jll package version number is not the same as the underlying artifact, even though I understand why it is necessary for them to not be the same. Will there be a lookup table in the README of each _jll github project page? Or will there be some other programmatic way to do it, which might even have Pkg support?

giordano:
> I'm still not clear on how package maintainers will be able to specify which artifact version they want if the _jll package version number is not the same as the underlying artifact, even though I understand why it is necessary for them to not be the same. Will there be a lookup table in the README of each _jll github project page? Or will there be some other programmatic way to do it, which might even have Pkg support?

We haven't worked this out yet because it isn't a trivial problem.  If you have some ideas please share in Yggdrasil.

galenlynch:
>Not really common, but not totally impossible.

Exactly, it seems like a pretty unlikely problem.

I think that this PR should be merged before there is both consensus about, and implementation of, and improved version system in Yggdrasil—which may take quite some time. The change here simply provides access to a function in the most recent version of fftw. I have updated the Project.toml for this package so that it requires that version of fftw (3.3.9), which should just work for the vast majority of users, and indeed worked for CI here. The potential downside of adding this function is that due to their local Pkg state, it's possible that a user will have a fftw library that reflects the state of fftw around Sep 17 2019, instead of the released 3.3.9 version, causing this new function to error. However, this Pkg state seems unlikely, even though it is possible. The fix for any affected users would be to simply run `pkg> update`, and possibly unpin their FFTW_jll, if they had it pinned.

Furthermore, it seems like the reason that FFTW_jll was shipping a pre-released version of fftw was to provide access to `fftw_threads_set_callback`, enabling mutlithreaded FFTW plans using Julia's partr threads. I see substantial performance improvements in DSP.jl's convolutions when using fftw's multithreaded transformations, but different sized convolutions seem to perform best with different numbers of threads (often one is best). While packages can change the number of threads being used by fftw with `set_num_threads`, doing so currently changes the global state of Julia since there is no way to see what that number was beforehand. This new function, `get_num_threads`, allows package authors to programatically change the number of fftw threads without changing global sate, and in a way would unlock the benefits of the (great) change that @stevengj made to enable fftw to use Julia's threads, even though that change is causing some consternation with package versions at the moment.

stevengj (approved):

src/fft.jl
```diff
@@ -176,6 +176,10 @@ end
     ccall((:fftwf_plan_with_nthreads,libfftw3f), Cvoid, (Int32,), nthreads)
 end
 
+function get_num_threads()
+    ccall((:fftw_planner_nthreads,libfftw3), Cint, ())
```
stevengj, line 180: shouldn't this have an `fftw_vendor` check?

galenlynch: Yeah it should

galenlynch:
I added the vendor check. 

It occurs to me that the way I would want to use this would benefit from support in this package. I would want to acquire the `fftwlock`, change the number of threads, make some plans, and then restore the original number of threads, and then release the `fftwlock`. To minimize version numbers I could add it to this PR, if you approve.

The function would look something like:
```julia
@exclusive function set_num_threads(f, nthreads::Integer)
    orig_nthreads = get_num_threads()
    try
        ccall((:fftw_plan_with_nthreads,libfftw3), Cvoid, (Int32,), nthreads)
        ccall((:fftwf_plan_with_nthreads,libfftw3f), Cvoid, (Int32,), nthreads)
        f()
    finally
        ccall((:fftw_plan_with_nthreads,libfftw3), Cvoid, (Int32,), orig_nthreads)
        ccall((:fftwf_plan_with_nthreads,libfftw3f), Cvoid, (Int32,), orig_nthreads)
    end
end
```

and would enable planning like:
```julia
plan = FFTW.set_num_threads(4) do
    plan_rfft(...)
end
```

The function argument would have to be careful to avoid a context switch, e.g. no IO. However, it seems like you would want to hold the lock while making plans to ensure that the plans were made with the intended number of threads.

stevengj:
> The function argument would have to be careful to avoid a context switch, e.g. no IO.

Why?

galenlynch:
Yeah I guess it wouldn't be a big deal for the thread to block on io with the lock.

stevengj:
Yes, adding that `set_num_threads` method seems reasonable.

(You wouldn't need to `ccall` though — just call `set_num_threads(::Integer)`.)

stevengj:
Or we could add an `nthreads` keyword argument to the `plan_fft` functions.

galenlynch:
>You wouldn't need to ccall though — just call set_num_threads(::Integer)

Wouldn't `set_num_threads(::Integer)` try to acquire the lock again?

>Or we could add an nthreads keyword argument to the plan_fft functions.

That also sounds great

src/fft.jl
```diff
@@ -176,6 +176,14 @@ end
     ccall((:fftwf_plan_with_nthreads,libfftw3f), Cvoid, (Int32,), nthreads)
 end
 
+function get_num_threads()
+    @static if fftw_vendor == :fftw
+        ccall((:fftw_planner_nthreads,libfftw3), Cint, ())
+    else
+        error("This function is not currently supported by mkl")
```
stevengj, line 183: I'm worried about this, especially if we start using `get_num_threads` more extensively (e.g. to accept an `nthreads` keyword in planning).   Maybe for MKL we should implement the fallback solution of keeping our own global nthreads variable mirroring the one in FFTW/MKL.

galenlynch: That's a good idea. I could just use similar code as what was in my first PR. I just had a global that cached the last number of threads.

src/FFTW.jl
```diff
@@ -25,6 +25,7 @@ end
 
 # MKL provides its own FFTW
 const fftw_vendor = occursin("mkl_rt", libfftw3) ? :mkl : :fftw
+const fftw_last_num_threads = Ref(Cint(1))
```
stevengj, line 28: I would just call it `_last_num_threads` — start it with an underscore to emphasize that this is an internal thing that should never be used outside this module, and omit the `fftw` prefix since it isn't needed for FFTW itself and we are already in the module namespace.   And only define it `if fftw_vendor == :mkl`

src/fft.jl
```diff
@@ -171,16 +171,34 @@ end
...
 function get_num_threads()
     @static if fftw_vendor == :fftw
         ccall((:fftw_planner_nthreads,libfftw3), Cint, ())
     else
-        error("This function is not currently supported by mkl")
+        fftw_last_num_threads[]
+    end
+end
+
+@exclusive function set_num_threads(f, nthreads::Integer)
```
stevengj, line 195: `f::Function`?

galenlynch: Fixed

src/FFTW.jl
galenlynch: Done

galenlynch:
I have implemented the suggestions in your code reviews, as well as your suggestion to add a `num_threads` keyword to the `plan_...` functions. My approach to adding the `num_threads` keyword argument here is pretty naive, but I couldn't think of a better way of adding the keyword argument to all of these functions. I haven't yet added tests for this new keyword argument, since I'm not sure this is the way you'd want to proceed. If so, I'd be happy to try to add some tests.

galenlynch:
Gentle bump on this. Not sure what's left to be done but I would love to see this merged.

src/dct.jl
```diff
@@ -3,14 +3,14 @@
 # (This is part of the FFTW module.)
 
 """
-    plan_dct!(A [, dims [, flags [, timelimit]]])
+    plan_dct!(A [, dims [, flags [, timelimit [, num_threads]]]])
```
stevengj, line 6: Why aren't these keyword arguments?

src/FFTW.jl
```diff
@@ -25,6 +25,9 @@ end
 
 # MKL provides its own FFTW
 const fftw_vendor = occursin("mkl_rt", libfftw3) ? :mkl : :fftw
+@static if fftw_vendor == :mkl
```
stevengj, line 28: In the latest version these checks need to be `fftw_provider == "mkl"`

src/dct.jl
galenlynch: I was just following the existing style for the keyword arguments `flags` and `timelimit`. Happy to change the documentation style (and probably flags and timelimit too?) if you'd prefer.

tknopp:
bump. The depending package NFFT.jl is multi-threaded and it would be nice to chose the same number of threads all operations without changing global state.

IanButterworth:
Another friendly bump 🙏.
This would be nice to have.

giordano:
I merged `master` onto this branch, I hope @galenlynch won't mind :slightly_smiling_face:

PR #172, opened 2020-12-24
MassInstallAction: Install the CI workflow on this repository

timholy:
This pull request sets up the CI workflow on this repository.


timholy:
Someone who knows a bit more about this package should look this over. I'm not sure of whether the ENV variables in the .travis.yml script should be used in "matrix" format and in what stage of the build process.

In case it's not clear, the reason for this PR is that travis.org goes away in a week, and travis.com won't be free. So view this PR as a downpayment on the migration, but it will require the touch of someone who knows what s/he is doing.

giordano:
I'll do it later

giordano:
Ok, after a few attempts I think I reproduced the same setup as Travis before.  The failure on Windows + Julia nightly + MKL is unrelated (and very weird)

giordano:
Is there a built-in way to tell whether FFTW or MKL is being used?  I can think of showing the value of `FFTW.libfftw3`, but maybe there are better ways.  I find it quite strange that the [size of the artifacts is only ~3 MiB](https://github.com/JuliaMath/FFTW.jl/pull/172/checks?check_run_id=1606958589#step:4:14), which is the size of the FFTW artifact only.  MKL doesn't seem to be actually used?

timholy:
While I don't have the expertise to give this a serious review, this looks really nice to me. Thanks!

giordano:
Ok, I'm going to merge, if there are some details to fix we can do it afterwards, but at least we can keep CI going.

PR #176, opened 2021-01-01
minor tag for Reexport 1.0

ChrisRackauckas:


ChrisRackauckas:
The nightly MKL failure looks very unrelated so I'm going to go ahead with the tagging to unblock downstream.

PR #181, opened 2021-01-23
Add a few precompiles

timholy:


src/FFTW.jl
```diff
@@ -91,4 +91,9 @@ end
 include("fft.jl")
 include("dct.jl")
 
+if Base.VERSION >= v"1.4.2"
```
ararslan, line 94: Can you add a comment about why it's this particular version?

timholy, line 94:
```suggestion
if Base.VERSION >= v"1.4.2"     # avoid potential segfaults, see https://github.com/JuliaLang/julia/pull/35378
```

timholy: Done

ararslan (approved):

timholy:
Thanks for the review, @ararslan!

ararslan:
My pleasure :)

Issue #185, opened 2021-03-11
Docs stop being deployed?


navidcy:
In the Documentation I can't seem to be able to find the docs for the latest releases, e.g. v1 and later on.

<img width="380" alt="Screen Shot 2021-03-12 at 9 34 09 am" src="https://user-images.githubusercontent.com/7112768/110864407-2d739200-8316-11eb-9a8e-3f59de96aaee.png">

Have the docs stopped being build/deployed?

This line
```Julia
┌ Warning: removing `stable` and linking `stable` to `v0.2.4`.
158
└ @ Documenter ~/.julia/packages/Documenter/FuXcO/src/Documenter.jl:683
159
```
from the `julia-docdeploy` section in GitHub Actions log at https://github.com/JuliaMath/FFTW.jl/runs/1882720951?check_suite_focus=true#step:6:157 is a bit puzzling... Why linking to v0.2.4?

giordano:
Documentation for v1.3.2 is now available: https://juliamath.github.io/FFTW.jl/v1.3.2/.  The problems with Documenter setup were already fixed by #179.

PR #186, opened 2021-03-13
Adds version in deploydocs

navidcy:
This is an attempt to deal with #185.

Unfortunately I don't think that I can push a docs preview, even if I set `push_preview = true` inside `deploydocs()` function so I don't have a way to figure out if this indeed deals with #185.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 [![Coveralls](https://coveralls.io/repos/github/JuliaMath/FFTW.jl/badge.svg?branch=master)](https://coveralls.io/github/JuliaMath/FFTW.jl?branch=master)
 
 [![](https://img.shields.io/badge/docs-stable-blue.svg)](https://JuliaMath.github.io/FFTW.jl/stable)
-[![](https://img.shields.io/badge/docs-latest-blue.svg)](https://JuliaMath.github.io/FFTW.jl/latest)
+[![](https://img.shields.io/badge/docs-dev-blue.svg)](https://JuliaMath.github.io/FFTW.jl/dev)
 
 This package provides Julia bindings to the [FFTW](http://www.fftw.org/) library for
 fast Fourier transforms (FFTs), as well as functionality useful for signal processing.
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -12,6 +12,7 @@ makedocs(
 
 deploydocs(
     repo = "github.com/JuliaMath/FFTW.jl.git",
+    versions = ["stable" => "v^", "v#.#.#", "dev" => "dev"],
     target = "build",
     deps = nothing,
     make = nothing,
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -12,8 +12,5 @@ makedocs(
 
 deploydocs(
     repo = "github.com/JuliaMath/FFTW.jl.git",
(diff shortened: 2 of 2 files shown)


giordano:
As far as I can see, the issue is that deploy of documentation on a tag is skipped: https://github.com/JuliaMath/FFTW.jl/runs/1871754241?check_suite_focus=true#step:6:84  No idea why that happens though (and not sure this helps, I mean, I've never seen this flag used)

navidcy:
😕

navidcy:
> As far as I can see, the issue is that deploy of documentation on a tag is skipped: https://github.com/JuliaMath/FFTW.jl/runs/1871754241?check_suite_focus=true#step:6:84 No idea why that happens though (and not sure this PR helps, I mean, I've never seen this flag used)

is this related? I've never seen `julia=...` kwarg in `deplaydocs()`....
https://github.com/JuliaMath/FFTW.jl/blob/6ad135ab0be80399156c12b341c70dfbd3429d7f/docs/make.jl#L15

giordano:
That's an old option, I think it isn't needed anymore, but I don't know whether that's the culprit (and I'd find it very bizarre if that was the case).

I have no clue about where the message
```
[ Info: skipping docs deployment.
```
comes from.  Finding it can give us a hint of _why_ deployment is skipped.

navidcy:
Searching `skipping docs deployment` in all of Documenter.jl repo give out nothing! Bizarre.

giordano:
Also, I removed that line in https://github.com/JuliaMath/FFTW.jl/commit/2c20abe7098a2380fa37fe021b3d21fc593059ec

navidcy:
@giordano, if you have write access to the repo then I suggest you create a PR copy of mine, add `push_preview = true` in the `deploydocs` and see if the docs actually deploy.

giordano:
How that would help if the issue is only on tags?

navidcy:
> How that would help if the issue is only on tags?

Oh now I see what you meant above. You are claiming the issue is on *every* tag... (I though you only were trying to say that a particular tag had the issue...)

navidcy:
Although, @giordano, the latest tagged release was with Documenter v0.22.6 and not the later ^0.26... But I have no idea if that's the culprit. Perhaps it's time to ask mortenpi (I'm not tagging them here... I'll let you do that if you think that's in order at this point).

giordano:
I already asked for help on Slack

giordano:
> the latest tagged release was with Documenter v0.22.6

That was a useful reminder, thanks.  I believe the message was coming from https://github.com/JuliaDocs/Documenter.jl/blob/70893ea4c0b17451ff1450a537c2768d226393d3/src/Documenter.jl#L503-L507, but that's already fixed by https://github.com/JuliaMath/FFTW.jl/commit/2c20abe7098a2380fa37fe021b3d21fc593059ec.  So nothing to do, I guess?

navidcy:
Well, there is a way to put back all the non deployed docs. But I’m not sure it worths it. 

https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#Documentation-Versions

giordano:
Documentation for v1.3.2 is now available: https://juliamath.github.io/FFTW.jl/v1.3.2/.  I'm going to close this pull request

Issue #188, opened 2021-03-31
Eliminate build step, use Preferences


staticfloat:
Now that 1.6 is released, I would like to eliminate the `Pkg.build()` step within FFTW entirely, instead changing it to use [Preferences](https://github.com/JuliaPackaging/Preferences.jl/) to manage which backend its using.  I'd also like to explore ways that it can make optional use of MKL and not unconditionally install it as a dependency.

My proposed plan is to use `Preferences` to perform a lookup (at compile time) for the `provider` key and use that, instead of looking at the value of `libfftw3`.  So e.g. [this line](https://github.com/JuliaMath/FFTW.jl/blob/07eeabb543402725f1404b172a6284b4bd8faf0a/src/FFTW.jl#L27) would instead become something like:

```
const fftw_vendor = @load_preference("provider")
```

The `@load_preference` macro will record in the `.ji` file that it accessed the `provider` key at compile-time, and so Julia, when loading the `.ji` file, will check to see if the current `provider` key matches the one recorded in the `.ji` file and if they do not match, it will reject the `.ji` file and possibly recompile FFTW.  This allows for FFTW to use MKL in some projects, and to use FFTW_jll in others.

More importantly for me, it removes the need to write out state into the depot and eliminates the `Pkg.build()` step, which should help in some of our caching efforts in large-scale deployments (where we want to make the depot read-only, we want to transfer around only content-addressed chunks of data from machine to machine, etc... and FFTW needing to rebuild itself because of the `prefs/FFTW` file causes a cascade of recompilation).

This capability is, of course, 1.6+.  @stevengj how do you feel about branching and creating a 1.6+ release?

giordano:
Side note: making the package v1.6+ only would allow us to build the next version of the upstream `FFTW` library for `aarch64-apple-darwin` (#187)

stevengj:
Fine with me.

PR #189, opened 2021-04-09
Require Julia 1.6+, use Preferences, eliminate mutable state

staticfloat:
This adds support for `Preferences.jl` as the backing store for setting
a computation backend (currently only supports `fftw` and `mkl` as valid
choices).  The convenience method `set_provider!()` is given, but it is
little more than a wrapper around `Preferences.set_preferences!()`.
Users can alter their `LocalPreferences.toml` or `Project.toml` files by
hand as well; in the event that an illegal value is provided, an error
will be displayed and `fftw` will be chosen as the default value.

Preferences is a relatively new subsystem in Julia and it has some
lacking support in `Pkg`; so for now we are adding a separate
`test/Project.toml`, which will have preferences added into it on CI, as
there is a bug in propagating preferences when a single `Project.toml`
is used with extra dependencies for the tests.


staticfloat:
This closes https://github.com/JuliaMath/FFTW.jl/issues/188

X-ref: https://github.com/JuliaLang/Pkg.jl/issues/2500

stevengj (approved):

stevengj:
LGTM, assuming CI is okay…

giordano:
@staticfloat perhaps it'd also be useful to have some documentation about how to set the preference :slightly_smiling_face:

src/providers.jl
```diff
@@ -0,0 +1,80 @@
...
+        end
+    end
+end
+
+# If we're using MKL, load it in and set library paths appropriately.
+@static if fftw_provider == "mkl"
+    using IntelOpenMP_jll, MKL_jll
+    const libfftw3 = MKL_jll.libmkl_rt_path
+    const libfftw3f = libfftw3
+end
```
andreasnoack, line 80: This doesn't seem compatible with PackageCompiler. The paths will end up being compiled into the system image.

staticfloat: Correct, this should be done within `__init__()` if we need this to work with PackageCompiler.

andreasnoack: Generally we ned all packages to work with PackageCompiler and specifically we very much need this to work with PackageCompiler

PR #190, opened 2021-04-09
Run CI with Drone also on Alpine Linux

giordano:
Ref: https://discourse.julialang.org/t/testing-julia-packages-on-x86-64-alpine-linux-with-drone-ci/57313

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -1,18 +1,19 @@
-local Pipeline(os, arch, version) = {
+local Pipeline(os, arch, version, alpine=false) = {
     kind: "pipeline",
-    name: os+" - "+arch+" - Julia "+version,
+    name: os+" - "+arch+" - Julia "+version+(if alpine then " (Alpine)" else ""),
     platform: {
-	os: os,
-	arch: arch
+        os: os,
+        arch: arch
     },
     steps: [
-	{
-	    name: "build",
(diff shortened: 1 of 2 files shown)


.drone.jsonnet
```diff
@@ -1,18 +1,20 @@
...
-	    commands: [
-		"julia --project=. --check-bounds=yes --color=yes -e 'using InteractiveUtils; versioninfo(verbose=true); using Pkg; Pkg.test(coverage=true)'"
-	    ]
-	}
+        {
+            name: "Run tests",
+            image: "julia:"+version+(if alpine then "-alpine" else ""),
+            commands: [
+                "julia --project=. --check-bounds=yes --color=yes -e 'using InteractiveUtils; versioninfo(verbose=true)'",
+                "julia --project=. --check-bounds=yes --color=yes -e 'using Pkg; Pkg.build()'",
```
staticfloat, line 14: We don't need to call `Pkg.build()` anymore

staticfloat (changes requested):

.drone.jsonnet
giordano: Uh, right, I used my standard template for CI of julia packages on Drone with all the steps.  I removed (again) this one, thanks!

ararslan (approved):

PR #191, opened 2021-04-09
Simplify `test/Project.toml` and remove redundant deps in `Project.toml`

giordano:
CC: @staticfloat

### DIFF
--- a/src/providers.jl
+++ b/src/providers.jl
@@ -74,7 +74,7 @@ end
 
 # If we're using MKL, load it in and set library paths appropriately.
 @static if fftw_provider == "mkl"
-    using IntelOpenMP_jll, MKL_jll
+    using MKL_jll
     const libfftw3 = MKL_jll.libmkl_rt_path
     const libfftw3f = libfftw3
 end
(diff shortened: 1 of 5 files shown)


test/runtests.jl
```diff
@@ -5,6 +5,8 @@ using AbstractFFTs: Plan, plan_inv
 using Test
 using LinearAlgebra
 
+@info "FFTW provider: $(fftw_provider)"
```
giordano, line 8: Perhaps I'm missing it, but I couldn't find an easy way to tell what's the provider when running the tests

staticfloat: Yeah, I just know to look for the `DCT` tests, which only run on `fftw`.  But this is clearly better.  :)

src/providers.jl
```diff
@@ -74,7 +74,7 @@ end
 
 # If we're using MKL, load it in and set library paths appropriately.
 @static if fftw_provider == "mkl"
-    using IntelOpenMP_jll, MKL_jll
+    using MKL_jll
```
giordano, line 77: We had to explicitly load `IntelOpenMP_jll` because some old versions of `MKL_jll` didn't depend on it (see #127), but this had been fixed long ago by https://github.com/JuliaPackaging/Yggdrasil/pull/260

PR #197, opened 2021-04-19
prevent to library from being serialized at precompile time

KristofferC:
The interpolated variable `libfftw3` will evaluate at precompile time to the path to the local directory of the fftw library. This means that if you e.g. build a sysimage using FFTW and move it somewhere else that sysimage will not work properly. Instead, use `:libfftw3` which evaluates to the variable `libfftw3` which is set during runtime. This is likely what was originally intended.

cc @staticfloat @giordano @andreasnoack

### DIFF
--- a/src/fft.jl
+++ b/src/fft.jl
@@ -552,8 +552,8 @@ end
 
 # low-level FFTWPlan creation (for internal use in FFTW module)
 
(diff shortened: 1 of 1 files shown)


andreasnoack:
It would be great with a new release that includes this fix.

ararslan:
FYI master is now 1.6 only, so if we wanted a patch with this we'd need to do a backport, right?

andreasnoack:
1.6 is all I need

PR #205, opened 2021-06-01
Add a safepoint in the maybe_destroy_plan spin lock

### LINKED ISSUE #163: Deadlock bug in FFTW 1.2.4
I have observed a deadlock bug involving FFTW v1.2.4. I am using the same codebase as I did in https://github.com/JuliaMath/FFTW.jl/issues/161. It appears that the bug is affected by the number of threads I give Julia (still setting FFTW  to 1 thread, though). With 4 threads, the deadlock bug occurs and the program will freeze. With 16 threads, the bug does not occur (it could be that the likelihood is simply much smaller and so does not occur during the ~6 minute runtime). Furthermore, in my debugging, I found that I could add enough debugging print statements that slow down the program enough that the deadlock bug does not occur.

I am confident that the bug was introduced in v1.2.4 because I cannot reproduce the problem with v1.2.2 nor v1.2.3. I can reproduce the problem with Julia v1.4 and v1.5. I can reproduce this problem on multiple machines reliably.

Looking at the changes introduced in v1.2.4, could it be that this loop is waiting indefinitely? https://github.com/JuliaMath/FFTW.jl/compare/v1.2.3...v1.2.4

Edit: It may also be worth noting that when the deadlock occurs, all threads appear to be busy. The CPU usage on the machine jumps to exactly the correct percentage for the number of threads I give Julia (e.g., on an 8 thread machine with 4 Julia threads, the CPU usage jumps to 50% and stays there indefinitely).

Octogonapus:
This PR adds a safepoint in the `maybe_destroy_plan` spin lock to permit forward progress in that spin lock.
Without it, the spin lock blocks forward progress and can ultimately deadlock the program.
This was implemented according to [this observation](https://github.com/JuliaMath/FFTW.jl/issues/163#issuecomment-852444447).
I have tested this locally by running the minimum reproducible example; I can't get the program to deadlock any more.

I would like to write an automated test for this, but I can't think of one that would be reliable and run in a timely manner. My best method of testing this is running the minimal example code locally repeatedly and seeing if the program deadlocks.

Fixes #163.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FFTW"
 uuid = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
-version = "1.4.1"
+version = "1.4.2"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
--- a/src/fft.jl
+++ b/src/fft.jl
@@ -304,7 +304,11 @@ function maybe_destroy_plan(plan::FFTWPlan)
     # since task switches aren't permitted in finalizers.   This has suboptimal efficiency,
     # but we shouldn't waste too many cycles since destroying plans is quick and contention
     # should be rare.
-    while !trylock(deferred_destroy_lock); end
+    while !trylock(deferred_destroy_lock)
+        # Need a safepoint in here because without it, this loop block forward progress in the GC and can deadlock
+        # the program.
+        GC.safepoint()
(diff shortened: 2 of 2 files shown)


stevengj:
@vtjnash, looks good?

vtjnash (approved):

Octogonapus:
@stevengj Anything else you want me to do on this PR so that it's RTM?

PR #207, opened 2021-06-16
Make FFTW relocatable for PackageCompiler

staticfloat:
Because we need to know the FFTW vendor pretty early (at precompile
time to generate code, and at runtime) we need to load the path during
build (which bakes it into the precompile image), however we need to
update that path on next load within `__init__()` in case we're a part
of PackageCompiler and we need to be able to move things around without
requiring a recompile.  So we change all accesses of `libfftw3{,f}` to
be through a `Ref{String}()`, then update that ref within `__init__()`.

In the future, accesses of `FFTW_jll.libfftw3` will be through a
function, rather than a direct variable access, and we'll be able to get
these things for free, but that's still a ways down the line.

### DIFF
--- a/src/FFTW.jl
+++ b/src/FFTW.jl
@@ -23,9 +23,17 @@ function __init__()
         Base.depwarn("JULIA_FFTW_PROVIDER is deprecated; use FFTW.set_provider!() instead", :JULIA_FFTW_PROVIDER)
     end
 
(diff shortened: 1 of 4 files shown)


stevengj (approved):

andreasnoack:
We've now been able to test this and it makes the package work with PackageCompiler.

PR #219, opened 2021-09-29
In-place `irfft`/`brfft`

BioTurboNick:
Based on #54 and my attempted reworking of it, this is a subset of the functionality just to support a drop-in replacement of `irfft!` and `brfft!` for `irfft` and `brfft`.

It uses the `PaddedRFFTArray` type internally and returns a reinterpreted view of the input array containing the output.

`rfft!` is trickier to support and I ran into some issues with it, so I've cut that out for the moment but it could be added later.


BioTurboNick:
In light of discussion on merging with #54 , closing this.

Issue #272, opened 2023-06-14
add rrules for `r2r`


vpuri3:

vpuri3:
how is the gradient computed for `plan_dct`, if there's not `rrule` dor `dct`??
```julia
using LinearAlgebra, FFTW, Zygote

x = rand(4)
C = plan_dct(x)

f(x) = C \ (C * x) |> norm
g(x) = x |> dct |> idct |> norm

@show Zygote.gradient(f, x) # ([0.7499995699183157, 0.5170775887690442, 0.3522881598130941, 0.2145331321046639],)
@show Zygote.gradient(g, x) # errors
```

```julia
julia> Zygote.gradient(f, x)                                                          
ERROR: Compiling Tuple{Type{FFTW.r2rFFTWPlan{Float64, Any, false, 1}}, Vector{Float64}, FFTW.FakeArray{Float64, 1}, UnitRange{Int64}, Int64, UInt32, Float64}: try/catch is n
ot supported.                                                                         
Refer to the Zygote documentation for fixes.                                          
https://fluxml.ai/Zygote.jl/latest/limitations                          
```

danielwe:
Cosigning this, having `AdjointStyle` and `adjoint_mul` for `r2r` plans would be great. I'm working out the one-dimensional case, but haven't quite wrapped my head around multidimensional FFTW yet

danielwe:
Here's an extremely rudimentary implementation of `adjoint_mul` for the 1d REDFT10, in case anyone finds it helpful as a starting point

```julia
using AbstractFFTs
using FFTW

struct R2RFFTAdjointStyle <: AbstractFFTs.AdjointStyle end

AbstractFFTs.AdjointStyle(::FFTW.r2rFFTWPlan) = R2RFFTAdjointStyle()

function AbstractFFTs.adjoint_mul(
    p::FFTW.r2rFFTWPlan{T}, x::AbstractVector{T}, ::R2RFFTAdjointStyle
) where {T}
    (length(p.kinds) == 1) || throw(ArgumentError("Multidimensional r2r transforms not yet supported"))
    (only(p.kinds) == 5) || throw(ArgumentError("r2r kinds other than REDFT10 not yet supported"))
    pinv = inv(p)
    unscaled_pinv = (pinv isa AbstractFFTs.ScaledPlan) ? pinv.p : pinv
    y = unscaled_pinv * x
    # REDFT10 is unitary except for the first row, so the unscaled inverse is its adjoint
[...]

PR #273, opened 2023-06-26
chain rules for DCT

vpuri3:
address https://github.com/JuliaMath/FFTW.jl/issues/272

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -9,11 +9,22 @@ LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MKL_jll = "856f044c-d86e-5d09-b602-aeab76dc8ba7"
 Preferences = "21216c6a-2e73-6563-6e65-726566657250"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
+Requires = "ae029012-a4dd-5104-9daa-d747884805df"
+
+[extensions]
+FFTWChainRulesCoreExt = "ChainRulesCore"
 
 [compat]
 AbstractFFTs = "1.0"
+ChainRulesCore = "1"
 FFTW_jll = "3.3.9"
 MKL_jll = "2019.0.117, 2020, 2021, 2022, 2023"
 Preferences = "1.2"
 Reexport = "0.2, 1.0"
 julia = "1.6"
+
+[extras]
+ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
+
+[weakdeps]
+ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/Project.toml
+++ b/Project.toml
@@ -9,7 +9,7 @@ LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MKL_jll = "856f044c-d86e-5d09-b602-aeab76dc8ba7"
 Preferences = "21216c6a-2e73-6563-6e65-726566657250"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
-Requires = "ae029012-a4dd-5104-9daa-d747884805df"
+ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 
 [extensions]
 FFTWChainRulesCoreExt = "ChainRulesCore"
--- a/ext/FFTWChainRulesCoreExt.jl
+++ b/ext/FFTWChainRulesCoreExt.jl
@@ -69,29 +69,4 @@ function ChainRulesCore.frule(Δ, ::typeof(r2r), x::AbstractArray, region...)
     return y, Δy
 end
 
-function ChainRulesCore.rrule(::typeof(r2r), x::AbstractArray, kinds, region...)
-    y = r2r(x, kinds, region...)
-    kinvs = if kinds isa Integer
-        FFTW.inv_kind[kinds]
-    else
-        Tuple(FFTW.inv_kind[k] for k in kinds)
-    end
-    project_x = ProjectTo(x)
-
-    function r2r_pullback(ȳ)
-        f̄ = NoTangent()
-        x̄ = project_x(r2r(unthunk(ȳ), kinvs, region...))
-        k̄ = NoTangent()
-        r̄ = NoTangent()
-
-        if isempty(region)
-            return f̄, x̄, k̄
-        else
-            return f̄, x̄, k̄, r̄
-        end
-    end
-
-    return y, r2r_pullback
-end
-
 end # module
--- a/ext/FFTWChainRulesCoreExt.jl
+++ b/ext/FFTWChainRulesCoreExt.jl
@@ -6,46 +6,28 @@ using ChainRulesCore
 
 # DCT/IDCT
 
-function ChainRulesCore.frule(Δ, ::typeof(dct), x::AbstractArray, region = 1:ndims(x))
-    Δx = Δ[2]
-    y = dct(x, region)
-    Δy = dct(Δx, region)
-    return y, Δy
-end
+for (fwd, bwd) in (
+    (dct, idct),
+    (idct, dct),
+)
+    function ChainRulesCore.frule(Δ, ::typeof(fwd), x::AbstractArray, region = 1:ndims(x))
+        Δx = Δ[2]
+        y = fwd(x, region)
+        Δy = fwd(Δx, region)
+        return y, Δy
+    end
 
-function ChainRulesCore.rrule(::typeof(dct), x::AbstractArray)
-    project_x = ProjectTo(x)
-    region = 1:ndims(x)
-    dct_pb(Δ) = NoTangent(), project_x(idct(unthunk(Δ), region))
-    return dct(x, region), dct_pb
-end
+    function ChainRulesCore.rrule(::typeof(fwd), x::AbstractArray)
+        project_x = ProjectTo(x)
+        dct_pb(Δ) = NoTangent(), project_x(bwd(unthunk(Δ)))
+        return fwd(x), dct_pb
+    end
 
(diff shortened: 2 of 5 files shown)


vpuri3:
This error is only happening with the MKL provider. With MKL, FFTW.jl doesn't even compile on my machine. Could be due to https://github.com/JuliaMath/FFTW.jl/commit/008bc5b0b80be127c304d5cf8943d3a657e1e4f7?

```julia
test_frule: idct on Array{Float64, 3},Int64: Error During Test at /home/runner/.julia/packages/ChainRulesTestUtils/C9L2i/src/testers.jl:123
  Got exception outside of a @test
  FFTW could not create plan
  Stacktrace:
    [1] error(s::String)
      @ Base ./error.jl:35
    [2] macro expansion
      @ FFTW ~/work/FFTW.jl/FFTW.jl/src/fft.jl:722 [inlined]
 ```

**Comment by vpuri3** [comment]
@devmotion, could you please review this? LMK if you want me to remove the `requires` part.

**Review by devmotion — COMMENTED** [review]
Can you explain the idea of the PR? The design goal was to define the differentiation rules in AbstractFFTs via adjoint plans. Downstream packages were supposed to implement this new adjoint interface.

**Comment by vpuri3** [comment]
Only the MKL tests are failing. Is there a regression with MKL?

**Comment by devmotion** [comment]
> Can you explain the idea of the PR? The design goal was to define the differentiation rules in AbstractFFTs via adjoint plans. Downstream packages were supposed to implement this new adjoint interface.

More concretely, https://github.com/JuliaMath/FFTW.jl/pull/249 outlines the intended approach. FFTW was not supposed to define custom rules.

**Comment by vpuri3** [comment]
This PR defines Chain Rules for `FFTW.dct, idct` since these methods are not defined in `AbstractFFTs`.

**Comment by devmotion** [comment]
Ref: https://github.com/JuliaMath/AbstractFFTs.jl/issues/56

**Comment by vpuri3** [comment]
With #249 , gradient computation would error for DCT/IDCT. This is because `dct(x)` is defined as `dct(x) = plan_dct(x) * x`. This is why we need an rrule for `dct`.

```julia
julia> using FFTW, Zygote                                                                                                                                                                                       
                                                                                                                                                                                                                
julia>                                                                                                                                                                                                          
julia> using LinearAlgebra, FFTW, Zygote                                                                                                                                                                        
                                                                                                                                                                                                                
julia> x = rand(4)                                                                                                                                                                                              
4-element Vector{Float64}:                                                                                                                                                                                      
 0.8692266334693106                                                                                                                                                                                             
 0.6938635624794242                                                                                                                                                                                             
 0.552208368655668                                                                                                                                                                                              
 0.9197557963740512                                                                                                                                                                                             
                                                                                                                                                                                                                
julia> f(x) = x |> dct |> idct |> norm                                                                                                                                                                          
f (generic function with 1 method)                                                                                                                                                                              
                                                                                                                                                                                                                
julia> f(x)                                                                                                                                                                                                     
1.5452787421840921                                                                                                                                                                                              
                                                                                                                                                                                                                
julia> Zygote.gradient(f, x)                                                                                                                                                                                    
ERROR: Compiling Tuple{Type{FFTW.r2rFFTWPlan{Float64, Any, false, 1}}, Vector{Float64}, FFTW.FakeArray{Float64, 1}, UnitRange{Int64}, Int64, UInt32, Float64}: try/catch is not supported.                      
Refer to the Zygote documentation for fixes.        
https://fluxml.ai/Zygote.jl/latest/limitations                                                                                                                                                                  
```

Project.toml
```diff
@@ -9,11 +9,22 @@ LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MKL_jll = "856f044c-d86e-5d09-b602-aeab76dc8ba7"
 Preferences = "21216c6a-2e73-6563-6e65-726566657250"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
+Requires = "ae029012-a4dd-5104-9daa-d747884805df"
```
devmotion, line 12: I think the ChainRules dependency should be a regular dependency on Julia < 1.9. AbstractFFTs depends on it anyway, so it is already an indirect dependency and the rules would always be loaded via Requires anyway - but due to the deficiencies of Requires they could not be precompiled.
```suggestion
ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
```

ext/FFTWChainRulesCoreExt.jl
```diff
@@ -0,0 +1,72 @@
+module FFTWChainRulesCoreExt
+
+using FFTW
+using FFTW: r2r
+using ChainRulesCore
+
+# DCT
+
+function ChainRulesCore.frule(Δ, ::typeof(dct), x::AbstractArray, region...)
```
devmotion, line 9: IMO the call sequences in FFTW are a bit weird - it seems at some point a missing `region` is set to `1:ndims(x)` but sometimes this only happens when the corresponding plan is created.... Since we don't want to support arbitrary arguments maybe just set
```suggestion
function ChainRulesCore.frule(Δ, ::typeof(dct), x::AbstractArray, region=1:ndims(x))
```
or use a default of `nothing` and branch in the function body?

ext/FFTWChainRulesCoreExt.jl
```diff
@@ -0,0 +1,72 @@
...
+    Δx = Δ[2]
+    y = dct(x, region...)
+    Δy = dct(Δx, region...)
+    return y, Δy
+end
+
+function ChainRulesCore.rrule(::typeof(dct), x::AbstractArray, region...)
```
devmotion, line 16: Same here, would have been nice to be able to just use `region` or `region=1:ndims(x)` - also to avoid the branch in the pullback below.

ext/FFTWChainRulesCoreExt.jl
```diff
@@ -0,0 +1,72 @@
...
+            return f̄, x̄, r̄
+        end
+    end
+
+    return y, idct_pullback
+end
+
+# R2R
+
+function ChainRulesCore.frule(Δ, ::typeof(r2r), x::AbstractArray, region...)
```
devmotion, line 65: It seems the `rrule` for `r2r` is missing?

src/FFTW.jl
```diff
@@ -16,6 +16,10 @@ export dct, idct, dct!, idct!, plan_dct, plan_idct, plan_dct!, plan_idct!
 
 include("providers.jl")
 
+@static if !isdefined(Base, :get_extension)
+    import Requires
+end
+
```
devmotion, line 22:
```suggestion
```

src/FFTW.jl
```diff
@@ -35,6 +39,12 @@ function __init__()
         libfftw3[] = MKL_jll.libmkl_rt_path
         libfftw3f[] = MKL_jll.libmkl_rt_path
     end
+
+    @static if !isdefined(Base, :get_extension)
+        Requires.@require ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4" begin
+            include("../ext/FFTWChainRulesCoreExt.jl")
+        end
+    end
```
devmotion, line 47: Just always load it outside of `__init__` if `!isdefined(Base, :get_extension)`.

test/runtests.jl
```diff
@@ -577,3 +577,47 @@ end
...
+                    test_frule(f, x, region)
+                    test_rrule(f, x, region)
+                end # for region
+            end # for x
+        end # for f
+    end
+
+    @testset "r2r" begin
+        for k in 4 #0:10
+            for x in (randn(3), )#randn(3, 4), randn(3, 4, 5))
```
devmotion, line 612: Something broken?

devmotion (commented):
Sorry, somehow I missed yesterday that the PR does not add rules for plans but for the `dct`, `idct` and `r2r` functions :facepalm: 

As long as they/their interface is not moved to AbstractFFTs, rules should be defined here :+1:

ext/FFTWChainRulesCoreExt.jl
vpuri3: done!

ext/FFTWChainRulesCoreExt.jl
vpuri3: You are right, the call sequence is a bit weird. I was attempting to get away with defining a single `rrule` for both `dct(x)` and `dct(x, region)`.

In FFTW.jl, if `dct(x)` were set to `dct(x, 1:ndims(x))`, then we could get away with defining an rrule on `dct(x, region)`. But `dct(x)` actually calls `plan_dct(x) * x`.

I tried your suggestion of setting `region = 1:ndims(x)` in the `rrule`, but then either `dct(x, region)` or `dct(x)` fail with below error depending on whether I return a cotangent for `region`.
```julia
test_rrule: dct on Matrix{Float64}: Test Failed at /Users/vp/.julia/packages/ChainRulesTestUtils/C9L2i/src/testers.jl:226                                                                                       
  Expression: length(ad_cotangents) == length(primals)                                                                                                                                                          
  Problem: The pullback should return 1 cotangent for the primal and each primal input.                                                                                                                         
   Evaluated: 3 == 2                                                                                                                                                                                            
```

vpuri3: It's easier to write different rules for `dct(x)`, `dct(x, region)`. That is done in https://github.com/JuliaMath/FFTW.jl/pull/273/commits/6b8d178afd368f1b5e5334b76e8b9b617fc3e6d6.

ext/FFTWChainRulesCoreExt.jl
vpuri3: The R2R transforms are not unitary. There is some scaling involved that depends on the kind of R2R transform. Because it looks like an involved task, I chose to skip that for now. I am happy to look into that in a separate PR

vpuri3:
@devmotion I've addressed all your comments. LMK if you have more questions

vpuri3:
@devmotion ping :)

Issue #133, opened 2019-12-18
Pkg.add("FFTW") shouldn't install MKL


antoine-levitt:
Since https://github.com/JuliaMath/FFTW.jl/commit/8a9197a08890ca18b3aeb8bdff568087026778d8, `]add FFTW` installs MKL. I'm not a free software absolutist, but I don't think adding such a core package (that used to be part of base) should download a non-free binary dependency. A much more practical concern is that MKL is a whopping 130mb, and my work connection sucks...

ararslan:
I definitely agree, but this is a difficult problem to address at the moment, since there is no support in the package manager for conditional dependencies.

antoine-levitt:
Given that seems to be still some time away, perhaps reverting to the previous state would be best? Even with a good connection the download still takes quite a bit of time (during which it is unclear what's happening)

ararslan:
The previous state was having MKL managed by Conda, which introduces a Python dependency. It sort of seems like there are no good options at the moment...

antoine-levitt:
I know very little about these things (and am eternally grateful to the poor souls that have to manage this) so sorry if I'm missing something, but isn't depending on python way better than depending on MKL? It's free software, it's more lightweight, it's shared by more packages, it runs on all sane platforms, and it's installed on most machines already.

Keno:
> isn't depending on python way better than depending on MKL

It was using Python to install MKL

antoine-levitt:
Not by default though? It's definitely the first time I've noticed `]add FFTW` downloads 130mb.

stevengj:
Previously it only installed Conda etcetera if FFTW.jl was configured to use MKL.  Now, because Pkg doesn't support optional dependencies, we must unconditionally download MKL_jll.

antoine-levitt:
So why not switch back to the old system until Pkg supports optional dependencies?

xgdgsc:
Please revert this. You are making my life with a bad network and AMD processors very hard.

KristofferC:
The artifact system already supports "lazy artifacts" that are only downloaded when the artifact is actually used.

giordano:
With latest release of `MKL_jll`, `v2019.0.117+2` (don't dare to install `v2019.0.117+1`, make sure to update the registry) the tarball of `MKL_jll` is not downloaded upon just installing the package, but only when doing `using MKL_jll`.

[...]

PR #278, opened 2023-09-28
Check and use only available FFTW libraries

### LINKED ISSUE #274: Cannot install FFTW on Apple M2-MAX chip, `libmkl_rt_path` not defined
Hi everyone,

I've got the following error when installing the "**FFTW**" package with Julia version **1.9.1** on **M2 Max** chip with **Ventura 13.4**

Any tips on resolving this?

 `using FFTW`
 │ Package FFTW not found, but a package named FFTW is available from a registry. 
 │ Install package?
 │   (@v1.9) pkg> add FFTW 
 └ (y/n/o) [y]: y
    Updating registry at `~/.julia/registries/General.toml`
   Resolving package versions...
    Updating `~/.julia/environments/v1.9/Project.toml`
  [7a1cc6ca] + FFTW v1.7.1
    Updating `~/.julia/environments/v1.9/Manifest.toml`
  [621f4979] + AbstractFFTs v1.3.2
  [7a1cc6ca] + FFTW v1.7.1
  [f5851436] + FFTW_jll v3.3.10+0
  [1d5cc7b8] + IntelOpenMP_jll v2023.1.0+0
  [856f044c] + MKL_jll v2023.1.0+0
[ Info: Precompiling FFTW [7a1cc6ca-52ef-59f5-83cd-3a7055c09341]
ERROR: LoadError: UndefVarError: `libmkl_rt_path` not defined
Stacktrace:
 [1] getproperty(x::Module, f::Symbol)
   @ Base ./Base.jl:31
 [2] top-level scope
   @ ~/.julia/packages/FFTW/HfEjB/src/providers.jl:86
 [3] include(mod::Module, _path::String)
   @ Base ./Base.jl:457
 [4] include(x::String)
   @ FFTW ~/.julia/packages/FFTW/HfEjB/src/FFTW.jl:1
 [5] top-level scope
   @ ~/.julia/packages/FFTW/HfEjB/src/FFTW.jl:17
 [6] include
   @ ./Base.jl:457 [inlined]
 [7] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, source::Nothing)
   @ Base ./loading.jl:2045
 [8] top-level scope
   @ stdin:3
in expression starting at /Users/.julia/packages/FFTW/HfEjB/src/providers.jl:84
in expression starting at /Users/.julia/packages/FFTW/HfEjB/src/FFTW.jl:1
in expression starting at stdin:3
ERROR: Failed to precompile FFTW [7a1cc6ca-52ef-59f5-83cd-3a7055c09341] to "/Users/.julia/compiled/v1.9/FFTW/jl_HmZkkA".
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] compilecache(pkg::Base.PkgId, path::String, internal_stderr::IO, internal_stdout::IO, keep_loaded_modules::Bool)
   @ Base ./loading.jl:2296
 [3] compilecache
   @ ./loading.jl:2163 [inlined]
 [4] _require(pkg::Base.PkgId, env::String)
   @ Base ./loading.jl:1805
 [5] _require_prelocked(uuidkey::Base.PkgId, env::String)
   @ Base ./loading.jl:1660
 [6] macro expansion
   @ ./loading.jl:1648 [inlined]
 [7] macro expansion
   @ ./lock.jl:267 [inlined]
 [8] require(into::Module, mod::Symbol)
   @ Base ./loading.jl:1611

devmotion:
I think it would be good to check if FFTW_jll and MKL_jll are actually available before trying to load them, as suggested in https://github.com/JuliaMath/FFTW.jl/issues/274#issuecomment-1618774373. I ran into this issue with a project that specified `provider = "mkl"` in a LocalPreferences.toml - precompilation will error on my MacBook since MKL is not available for Apple silicon. With this PR, FFTW will use FFTW_jll instead (if it's not available either an error is thrown) but present the user with the error message:
```julia
┌ FFTW [7a1cc6ca-52ef-59f5-83cd-3a7055c09341]
│  ┌ Error: Invalid provider setting "mkl"; valid settings include ["fftw"]
│  └ @ FFTW ~/.julia/dev/FFTW/src/providers.jl:23
└
```

Fixes https://github.com/JuliaMath/FFTW.jl/issues/274.

### DIFF
--- a/src/FFTW.jl
+++ b/src/FFTW.jl
@@ -11,6 +11,8 @@ import AbstractFFTs: Plan, ScaledPlan,
                      fftshift, ifftshift,
                      rfft_output_size, brfft_output_size,
                      plan_inv, normalization
+import FFTW_jll
+import MKL_jll
 
 export dct, idct, dct!, idct!, plan_dct, plan_idct, plan_dct!, plan_idct!
 
--- a/src/FFTW.jl
+++ b/src/FFTW.jl
@@ -11,8 +11,6 @@ import AbstractFFTs: Plan, ScaledPlan,
                      fftshift, ifftshift,
                      rfft_output_size, brfft_output_size,
                      plan_inv, normalization
-import FFTW_jll
-import MKL_jll
 
 export dct, idct, dct!, idct!, plan_dct, plan_idct, plan_dct!, plan_idct!
 
--- a/src/providers.jl
+++ b/src/providers.jl
@@ -1,10 +1,26 @@
-const valid_fftw_providers = if FFTW_jll.is_available() && MKL_jll.is_available()
-    ("fftw", "mkl")
-elseif FFTW_jll.is_available()
-    ("fftw",)
-elseif MKL_jll.is_available()
-    ("mkl",)
-else
+# Hardcoded list of supported platforms
+# In principle, we could check FFTW_jll.is_available() and MKL_jll.is_available()
+# but then we would have to load MKL_jll which we want to avoid (lazy artifacts!)
+const platforms_providers = Dict(
+    Base.BinaryPlatforms.Platform("aarch64", "macos") => ("fftw",),
+    Base.BinaryPlatforms.Platform("aarch64", "linux"; libc = "glibc") => ("fftw",),
+    Base.BinaryPlatforms.Platform("aarch64", "linux"; libc = "musl") => ("fftw",),
+    Base.BinaryPlatforms.Platform("armv6l", "linux"; libc = "glibc", call_abi = "eabihf") => ("fftw",),
+    Base.BinaryPlatforms.Platform("armv6l", "linux"; libc = "musl", call_abi = "eabihf") => ("fftw",),
+    Base.BinaryPlatforms.Platform("armv7l", "linux"; libc = "glibc", call_abi = "eabihf") => ("fftw",),
+    Base.BinaryPlatforms.Platform("armv7l", "linux"; libc = "musl", call_abi = "eabihf") => ("fftw",),
+    Base.BinaryPlatforms.Platform("i686", "linux"; libc = "glibc") => ("fftw", "mkl"),
+    Base.BinaryPlatforms.Platform("i686", "linux"; libc = "musl") => ("fftw",),
+    Base.BinaryPlatforms.Platform("i686", "windows") => ("fftw", "mkl"),
+    Base.BinaryPlatforms.Platform("powerpc64le", "linux"; libc = "glibc") => ("fftw",),
+    Base.BinaryPlatforms.Platform("x86_64", "macos") => ("fftw", "mkl"),
+    Base.BinaryPlatforms.Platform("x86_64", "linux"; libc = "glibc") => ("fftw",),
+    Base.BinaryPlatforms.Platform("x86_64", "linux"; libc = "musl") => ("fftw",),
+    Base.BinaryPlatforms.Platform("x86_64", "freebsd") => ("fftw",),
+    Base.BinaryPlatforms.Platform("x86_64", "windows") => ("fftw", "mkl"),
+)
+const valid_fftw_providers = Base.BinaryPlatforms.select_platform(platforms_providers, Base.BinaryPlatforms.HostPlatform())
+if valid_fftw_providers === nothing
     error("no valid FFTW library available")
 end
 
@@ -56,6 +72,12 @@ end
 
 # If we're using fftw_jll, load it in
 @static if fftw_provider == "fftw"
+    import FFTW_jll
+    if !FFTW_jll.is_available()
+        # more descriptive error message if FFTW is not available
+        # (should not be possible to reach this)
+        error("FFTW library cannot be loaded: please switch to the `mkl` provider for FFTW.jl")
+    end
     libfftw3[] = FFTW_jll.libfftw3_path
     libfftw3f[] = FFTW_jll.libfftw3f_path
 
@@ -90,6 +112,12 @@ end
 
 # If we're using MKL, load it in and set library paths appropriately.
 @static if fftw_provider == "mkl"
+    import MKL_jll
+    if !MKL_jll.is_available()
+        # more descriptive error message if MKL is not available
+        # (should not be possible to reach this)
+        error("MKL library cannot be loaded: please switch to the `fftw` provider for FFTW.jl")
+    end
     libfftw3[] = MKL_jll.libmkl_rt_path
     libfftw3f[] = MKL_jll.libmkl_rt_path
     const _last_num_threads = Ref(Cint(1))
--- a/src/providers.jl
+++ b/src/providers.jl
@@ -1,17 +1,26 @@
+const valid_fftw_providers = if FFTW_jll.is_available() && MKL_jll.is_available()
+    ("fftw", "mkl")
+elseif FFTW_jll.is_available()
+    ("fftw",)
+elseif MKL_jll.is_available()
+    ("mkl",)
+else
+    error("no valid FFTW library available")
+end
 
 function get_provider()
     # Note: we CANNOT do something like have the `default` value be `get(ENV, "JULIA_FFTW_PROVIDER", "fftw")` here.
     # This is because the only way the Julia knows that a default has changed is if the values on-disk change; so
     # if your "default" value can be changed from the outside, you quickly run into cache invalidation issues.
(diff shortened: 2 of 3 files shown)


src/FFTW.jl
```diff
@@ -11,6 +11,8 @@ import AbstractFFTs: Plan, ScaledPlan,
                      fftshift, ifftshift,
                      rfft_output_size, brfft_output_size,
                      plan_inv, normalization
+import FFTW_jll
+import MKL_jll
```
giordano, line 15: This will unconditionally force installation of MKL in all cases.

devmotion: Ah I was worried something like this would happen 😅

devmotion: So installation only happens when importing a jll library? Isn't it installed anyway since it's a package dependency? Or is this a MKL-specific thing?

src/providers.jl
```diff
@@ -82,7 +90,6 @@ end
 
 # If we're using MKL, load it in and set library paths appropriately.
 @static if fftw_provider == "mkl"
-    import MKL_jll
```
giordano, line 85: Why not doing the test `is_available` here?  People are not going to be happy to having to unconditionally download MKL, that was the whole point of #133

src/FFTW.jl
giordano: MKL_jll is lazy, which means it's installed only on `using`/`import`.

src/providers.jl
devmotion: I was not aware of this issue. Mainly, I thought it would be more user-friendly to figure out the supported libraries in advance, before trying to set them. Only moving the check here won't help with my use case and the linked issue above.

devmotion: Is there any other way to check whether a library can be loaded apart from `..._jll.is_available()`? Maybe checking the system architecture is an OKish workaround?

giordano: > Is there any other way to check whether a library can be loaded apart from ..._jll.is_available()?

No.

> Maybe checking the system architecture is an OKish workaround?

That's going to hard-code answers which may potentially change over time.

giordano: >  Only moving the check here won't help with [...] the linked issue above.

Why not?  Capture the fact that the artifact isn't available and throw a helpful error message to explain what to do, instead of getting a cryptic `UndefVarError` that some users don't know what to do with.

devmotion: Yeah, switching to the supported default provider was the main point - changing the LocalPreferences.toml file manually on my end creates git conflicts when moving branches and pulling all the time (and if I forget to adjust it, compilation will fail). To some extent that's already done in the current code: If you specify an invalid provider other than `fftw` and `mkl`, FFTW will just display an `@error` and switch to the default provider (which is fixed to `fftw` currently). So basically I'm in the same situation, on my computer `mkl` is also an invalid provider so I would like to switch to the default one. And I guess you could hypothetically think about an analogous scenario where `fftw` is invalid and you would like to switch to the `mkl` as default provider.

devmotion: Displaying a more descriptive error message is an improvement but I think a solution that just defines the (in)valid providers correctly would be much more convenient.

devmotion: > That's going to hard-code answers which may potentially change over time.

Yes, a bit annoying. But since I didn't see any other alternative to define the valid providers correctly, I switched to hardcoding the supported platforms.

I guess the supported platforms could always be updated easily - and probably supported platforms for MKL won't change much at least? If we're more ambitious, probably we could even automatically extract it from the Artifacts.toml files in FFTW_jll and MKL_jll.

src/providers.jl
```diff
@@ -1,17 +1,42 @@
...
+    Base.BinaryPlatforms.Platform("i686", "linux"; libc = "musl") => ("fftw",),
+    Base.BinaryPlatforms.Platform("i686", "windows") => ("fftw", "mkl"),
+    Base.BinaryPlatforms.Platform("powerpc64le", "linux"; libc = "glibc") => ("fftw",),
+    Base.BinaryPlatforms.Platform("x86_64", "macos") => ("fftw", "mkl"),
+    Base.BinaryPlatforms.Platform("x86_64", "linux"; libc = "glibc") => ("fftw",),
+    Base.BinaryPlatforms.Platform("x86_64", "linux"; libc = "musl") => ("fftw",),
+    Base.BinaryPlatforms.Platform("x86_64", "freebsd") => ("fftw",),
+    Base.BinaryPlatforms.Platform("x86_64", "windows") => ("fftw", "mkl"),
+)
+const valid_fftw_providers = Base.BinaryPlatforms.select_platform(platforms_providers, Base.BinaryPlatforms.HostPlatform())
```
giordano, line 22: If you do something like
```julia
const valid_fftw_providers = let
    platforms_providers = Dict(...)
    Base.BinaryPlatforms.select_platform(platforms_providers, Base.BinaryPlatforms.HostPlatform())
end
```
you avoid storing in the compile cache a useless dictionary.

src/providers.jl
```diff
@@ -48,6 +73,11 @@ end
 # If we're using fftw_jll, load it in
 @static if fftw_provider == "fftw"
     import FFTW_jll
+    if !FFTW_jll.is_available()
+        # more descriptive error message if FFTW is not available
+        # (should not be possible to reach this)
+        error("FFTW library cannot be loaded: please switch to the `mkl` provider for FFTW.jl")
```
giordano, line 79: Maybe provide a more helpful error message to explain how to to do this?  Most users won't have a clue of what this is about.

src/providers.jl
```diff
@@ -1,17 +1,44 @@
...
+        Base.BinaryPlatforms.Platform("aarch64", "linux"; libc = "musl") => ("fftw",),
+        Base.BinaryPlatforms.Platform("armv6l", "linux"; libc = "glibc", call_abi = "eabihf") => ("fftw",),
+        Base.BinaryPlatforms.Platform("armv6l", "linux"; libc = "musl", call_abi = "eabihf") => ("fftw",),
+        Base.BinaryPlatforms.Platform("armv7l", "linux"; libc = "glibc", call_abi = "eabihf") => ("fftw",),
+        Base.BinaryPlatforms.Platform("armv7l", "linux"; libc = "musl", call_abi = "eabihf") => ("fftw",),
+        Base.BinaryPlatforms.Platform("i686", "linux"; libc = "glibc") => ("fftw", "mkl"),
+        Base.BinaryPlatforms.Platform("i686", "linux"; libc = "musl") => ("fftw",),
+        Base.BinaryPlatforms.Platform("i686", "windows") => ("fftw", "mkl"),
+        Base.BinaryPlatforms.Platform("powerpc64le", "linux"; libc = "glibc") => ("fftw",),
+        Base.BinaryPlatforms.Platform("x86_64", "macos") => ("fftw", "mkl"),
```
giordano, line 17: #280 shows that the hardcoded list a bad idea: there's no MKL 2024 build for this platform.

src/providers.jl
giordano: > I guess the supported platforms could always be updated easily - and probably supported platforms for MKL won't change much at least?

It did change.

src/providers.jl
giordano: But why not removing the automagic fallback, and only do what the user asks, and loudly and clearly fail if the artifact isn't available? In that case you can use is_available safely, since the user explicitly requested it

devmotion: Because the motivation for the PR and the problem I was trying to solve was that it failed. As a user I don't want to run into an error, I just want to use the provider that is available on my platform.

devmotion: See e.g. also https://julialang.slack.com/archives/C6A044SQH/p1701979919655409

devmotion: To clarify a bit more: In my case the MKL provider was chosen as the default for me by a package I was using, but in principle it could also be the other way around (that the currently hardcoded default FFTW_jll provider does not work but MKL_jll does). I think it would be less surprising and more user-friendly if the default provider would not be hard-coded to e.g. FFTW_jll but that it is checked which providers are actually available for the user and the default one is chosen as the first one among those. I don't think one should override explicit user settings but I think the default should work (and the error messages/list of valid providers should match the ones that are actually available).

PR #286, opened 2023-12-10
CI: for the macOS CI jobs with the MKL provider, force MKL_jll 2023 to be installed

DilumAluthge:
In MKL 2024, Intel dropped support for macOS.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -49,6 +49,7 @@ jobs:
             arch: x86
           - provider: 'mkl'
             threads: '2'
+
     steps:
       - uses: actions/checkout@v3
       - uses: julia-actions/setup-julia@v1
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -49,10 +49,6 @@ jobs:
             arch: x86
           - provider: 'mkl'
             threads: '2'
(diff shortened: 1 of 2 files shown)


giordano:
Use MKL 2023 for testing those jobs, so we do something actually useful?

DilumAluthge:
> Use MKL 2023 for testing those jobs, so we do something actually useful?

Done!

.github/workflows/CI.yml
```diff
@@ -57,6 +57,11 @@ jobs:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
       - uses: julia-actions/cache@v1
+      # In MKL 2024, Intel dropped support for macOS.
+      # So, in CI, for the macOS jobs, we force MKL 2023 to be installed (instead of
+      # MKL 2024).
+      - run: julia .ci/macos_mkl_2023.jl
```
giordano, line 63: Wouldn't `Pkg.add(; name="MKL_jll"; version="2023")` be enough?

DilumAluthge: There are test dependencies, so in theory the `Pkg.test()` re-resolve can do whatever it wants.

ericphanson:
What about using `Pkg.add` then `Pkg.pin` so it can't resolve? (@palday's idea in an unrelated repo)

PR #296, opened 2024-02-23
Update CI.yml for Apple Silicon

ViralBShah:
Enabling Apple Silicon is a bit messy because of the cross product of all platforms. It's good to be able to test it though, and happy to improve the script if anyone has suggestions for simplification.

Is it useful to run all the tests on Julia nightly? I feel we can save CI time by removing nightly.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -61,15 +61,16 @@ jobs:
             arch: x86
           - os: macOS-13
             arch: aarch64
+          - os: macOS-13
+            version: '1.6'
+            provider: 'mkl'
           - os: macOS-14
             arch: x86
           - os: macOS-14
             arch: x64
           - os: macOS-14
-            arch: aarch64
             version: '1.6'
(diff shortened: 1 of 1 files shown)


ViralBShah:
The failing CI case is because the runner is unable to download MKL correctly (despite several attempts).

ViralBShah:
@DilumAluthge Any ideas how to get the failing action to pass? It can't download MKL_jll repeatedly. Perhaps safe to merge?

ViralBShah:
@ararslan Would it be ok to disable nightly in the CI?

ararslan:
> @ararslan Would it be ok to disable nightly in the CI?

Sorry, I just saw this. Perhaps we could simplify to only a single nightly job, e.g. just run on Linux x64? Testing on nightly can be useful to ensure upstream changes don't affect the package in unexpected ways but IMO a single platform would suffice.

ViralBShah:
Yes makes sense. I'll add that shortly.

PR #298, opened 2024-03-01
fix bug in plan_r2r! for num_threads != nothing

### LINKED ISSUE #297: Bug in plan_r2r! leads to false results for multiple threads
There is a typo in the code for creating a plan_r2r! with numthreads != nothing.
The function returns an out of place plan_r2r in this case (the plan_r2r is missing the ! )

```
function plan_r2r!(X::StridedArray{T,N}, kinds, region;
                   flags::Integer=ESTIMATE,
                   timelimit::Real=NO_TIMELIMIT,
                   num_threads::Union{Nothing, Integer} = nothing) where {T<:fftwNumber,N}
    if num_threads !== nothing
        plan = set_num_threads(num_threads) do
            plan_r2r(X, kinds, region; flags = flags, timelimit = timelimit)
        end
        return plan
    end
    r2rFFTWPlan{T,Any,true,N}(X, X, region, kinds, flags, timelimit)
end
```

If the plan is now out of place this silently breaks code relying on  P*x  acting in place on x since the out of place plan now does not modify x.

perhaps this is additional motivation to require lmul!(P,x) syntax for applying in place plans

stevengj:
Fixes #297

### DIFF
--- a/src/fft.jl
+++ b/src/fft.jl
@@ -987,7 +987,7 @@ function plan_r2r!(X::StridedArray{T,N}, kinds, region;
                    num_threads::Union{Nothing, Integer} = nothing) where {T<:fftwNumber,N}
     if num_threads !== nothing
         plan = set_num_threads(num_threads) do
-            plan_r2r(X, kinds, region; flags = flags, timelimit = timelimit)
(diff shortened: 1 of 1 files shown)


ViralBShah:
Merge?

PR #299, opened 2024-03-25
Move MKL to an extension on 1.9+

palday:
I've marked this as a breaking change because it requires explicitly loading MKL_jll (or MKL) to enable the support on 1.9+.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,12 +1,11 @@
 name = "FFTW"
 uuid = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
-version = "1.8"
+version = "2.0.0"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
 FFTW_jll = "f5851436-0d7a-5f13-b9de-f02708fd171a"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
-MKL_jll = "856f044c-d86e-5d09-b602-aeab76dc8ba7"
 Preferences = "21216c6a-2e73-6563-6e65-726566657250"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 
@@ -18,3 +17,12 @@ MKL_jll = "2019.0.117, 2020, 2021, 2022, 2023, 2024"
 Preferences = "1.2"
 Reexport = "0.2, 1.0"
 julia = "1.6"
+
+[extensions]
+FFTWMKLExt = ["MKL_jll"]
+
+[weakdeps]
+MKL_jll = "856f044c-d86e-5d09-b602-aeab76dc8ba7"
+
+[extras]
+MKL_jll = "856f044c-d86e-5d09-b602-aeab76dc8ba7"
--- a/Project.toml
+++ b/Project.toml
@@ -21,8 +21,8 @@ julia = "1.6"
 [extensions]
 FFTWMKLExt = ["MKL_jll"]
 
-[weakdeps]
+[extras]
 MKL_jll = "856f044c-d86e-5d09-b602-aeab76dc8ba7"
 
-[extras]
+[weakdeps]
 MKL_jll = "856f044c-d86e-5d09-b602-aeab76dc8ba7"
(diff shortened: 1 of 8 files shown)


stevengj:
Why is this a breaking change?  At worst, someone currently using the MKL backend might use the FFTW backend instead, which could change the performance but not the API.

giordano:
I can't have a look right now but I have a couple of questions:

* you can't control whether other packages in the environment load MKL or not, does it mean other packages may force you to use MKL as backend for FFTW.jj even if you didn't want to?
* what happens on platforms where MKL isn't available? Does loading the no-op MKL wreak FFTW.jl?

palday:
> Why is this a breaking change? At worst, someone currently using the MKL backend might use the FFTW backend instead, which could change the performance but not the API.

@stevengj The current logic (which I just copied over) doesn't fallback to the FFTW backend when MKL isn't available and the preference is set to MKL. Since MKL now has to be explicitly loaded to be available, this seemed like a breaking behavior to me. I can also change it to be a minor version bump.


 

> I can't have a look right now but I have a couple of questions:
> 
>     * you can't control whether other packages in the environment load MKL or not, does it mean other packages may force you to use MKL as backend for FFTW.jj even if you didn't want to?

No, the choice of backend is a user-set preference.

(Also, I view packages loading MKL as a form of piracy, at a much lower level than most people think of!)

> 
>     * what happens on platforms where MKL isn't available? Does loading the no-op MKL wreak FFTW.jl?

No, again because it still depends on the user preference. If MKL is loaded but the user preference is (the default) FFTW, then the package extension is also essentially a no-op (see the `@static` guards).

palday:
@stevengj Maybe the solution is to unconditionally set things to the FFTW backend in the package and only conditionally update for the `mkl` preference in the extension. Then there things still work even when MKL isn't loaded.

palday:
FWIW I suspect setting the backend to `mkl` in the current release on a platform without MKL support (e.g. Apple silicon) would result in breakage. So I guess one could argue that the current changes aren't breaking enough to warrant to warrant a breaking release.

stevengj:
What is the advantage of making this an extension over the current approach?

palday:
Storage space, download times (which matters more for CI) and unnecessary compat restrictions.

giordano:
> Setting the backend to mkl in the current release on a platform without MKL support (e.g. Apple silicon) would result in breakage.

But hopefully one doesn't purposefully do something which would just not work. Instead, you can't control at all whether the MKL_jll package just happens to be loaded (and it'd be No-op on platforms where MKL isn't available). To me the latter looks much more troublesome, as it's outside of user's control.

giordano:
Also, how do you *control* what's the FFT backend if loading another package toggle that? This mechanism looks very problematic to me.

palday:
> Also, how do you _control_ what's the FFT backend if loading another package toggle that? This mechanism looks very problematic to me.

@giordano Again, the user controls which backend to use via their preferences. The extension merely enables an additional backend _as an option_. If the user has not set that option, then it is not used. On the other hand, if a user wants to use a different backend, I think it's fair that they are required to load that backend, i.e. opt in fully. They still control whether or not to set the MKL preference and whether or not to load MKL. With this PR, they can load MKL without using it as a backend if they just want MKL for BLAS, but they can't use MKL as the backend without loading it. The user is in *more* control, not less.

stevengj:
What if the user has set the `mkl'` preference for their own usage, in which case they know to `import MKL_jll` before importing `FFTW`, but then they use some package `X` that (unknown to the user) also imports `FFTW`.   Then, when the user does `using X`, it will break, because `X` didn't know to `import MKL_jll`?

This doesn't seem very composable.

palday:
> What if the user has set the `mkl'` preference for their own usage, in which case they know to `import MKL_jll` before importing `FFTW`, but then they use some package `X` that (unknown to the user) also imports `FFTW`. Then, when the user does `using X`, it will break, because `X` didn't know to `import MKL_jll`?
> 
> This doesn't seem very composable.

That's a very good point for global preferences. For project specific preferences, it seems reasonable for the user to only set MKL as the backend if they load it. But I don't know what to do about the global preference thing and that might be the best argument against this change.

PR #307, opened 2024-10-02
[Enzyme] Mark assert applicable as non-differentiable

wsmoses:


wsmoses:
@stevengj mind giving this a review?

wsmoses:
It doesn’t yet differentiate it fully, but this is a necessary pre requisite

giordano:
I suspect this needs to add Enzyme to the extras to make Pkg happy. But also, might be easier to move the extension to Enzyme? Pkgextensions are symmetrical, it doesn't really matter which of the two packages hosts them.

giordano (approved):

giordano:
This is relevant: https://discourse.julialang.org/t/autodifferentiation-with-fft-with-enzyme/128674/2. CC @danielwe

PR #308, opened 2024-10-09
Improve type inference

topolarity:
This makes the following example compile with `--trim`:
```julia
using FFTW

Base.@ccallable function main() :: Cint
    v = fft([0; 1; 2; 1])
    for elem in v
        println(Core.stdout, elem)
    end
    return 0
end
```
```
$ ./fftw
4.0 + 0.0im
-2.0 + 0.0im
0.0 + 0.0im
-2.0 + 0.0im
```

### DIFF
--- a/src/fft.jl
+++ b/src/fft.jl
@@ -319,9 +319,9 @@ unsafe_convert(::Type{PlanPtr}, p::FFTWPlan) = p.plan
 # This is accomplished by the maybe_destroy_plan function, which is used as the plan finalizer.
 
 # these functions should only be called while the fftwlock is held
-unsafe_destroy_plan(plan::FFTWPlan{<:fftwDouble}) =
+unsafe_destroy_plan(@nospecialize(plan::FFTWPlan{<:fftwDouble})) =
     ccall((:fftw_destroy_plan,libfftw3[]), Cvoid, (PlanPtr,), plan)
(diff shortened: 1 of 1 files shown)


src/fft.jl
```diff
@@ -335,7 +335,12 @@ function destroy_deferred()
         # we'll do nothing (the other function will eventually run destroy_deferred).
         if !isempty(deferred_destroy_plans) && trylock(fftwlock)
             try
-                foreach(unsafe_destroy_plan, deferred_destroy_plans)
+                @static if Base.VERSION >= v"1.9"
+                    @inline foreach(unsafe_destroy_plan, deferred_destroy_plans)
```
stevengj, line 339: Why is inlining needed here?

src/fft.jl
```diff
@@ -319,9 +319,9 @@ unsafe_convert(::Type{PlanPtr}, p::FFTWPlan) = p.plan
 # This is accomplished by the maybe_destroy_plan function, which is used as the plan finalizer.
 
 # these functions should only be called while the fftwlock is held
-unsafe_destroy_plan(plan::FFTWPlan{<:fftwDouble}) =
+unsafe_destroy_plan(@nospecialize(plan::FFTWPlan{<:fftwDouble})) =
```
stevengj, line 322: What problem does this solve, exactly?

PR #311, opened 2024-12-13
WIP: ForwardDiff extension

dlfivefifty:
This moves over code from FastTransformsForwardDiff.jl, and goes with https://github.com/JuliaMath/AbstractFFTs.jl/pull/138


dlfivefifty:
We first need to merge the AbstractFFTs.jl extension. That was held up because there's no test plans for real transforms (like DCT). Also I got busy.

I can probably complete this next month.

PR #314, opened 2025-01-22
allow MKL_jll version 2025

ranocha:
A new version has been released; see https://github.com/JuliaBinaryWrappers/MKL_jll.jl. We observe some issues resulting from the package manager deciding to use an old version of FFTW.jl that does not impose restrictions on MKL_jll so that the 2025 version can be installed.


giordano:
You need to remove the x86 jobs with MKL though.

ranocha:
Like this, @giordano?

.github/workflows/CI.yml
```diff
@@ -52,6 +52,10 @@ jobs:
             arch: x86
           - provider: 'mkl'
             threads: '2'
+          # x86 has issues with MKL 2025
```
giordano, line 55:
```suggestion
          # MKL 2025 doesn't support i686
```

.github/workflows/CI.yml
```diff
@@ -52,6 +52,10 @@ jobs:
...
+          - os: windows-latest
+            provider: 'mkl'
```
giordano, line 57:
```suggestion
          - provider: 'mkl'
```

giordano: And delete https://github.com/JuliaMath/FFTW.jl/blob/7e845933e80e70b6a06576e7abcc0206c9ff51f7/.github/workflows/CI.yml#L49-L54

giordano:
https://github.com/JuliaMath/FFTW.jl/pull/314#discussion_r1925753010

> And delete
> 
> https://github.com/JuliaMath/FFTW.jl/blob/7e845933e80e70b6a06576e7abcc0206c9ff51f7/.github/workflows/CI.yml#L49-L54

ranocha:
Sorry, I didn't see that comment earlier - should be fixed now

PR #322, opened 2025-08-20
Add external links to the documentation and update Documenter compat

kunzaatko:
- **fix(docs): Update compat bound for Documenter and dev the package**
- **feat(docs): Add the DocumenterInterLinks package for external documentation links**
- **feat(docs): Use `@extref` for external references**
- **fix(docs): Add `warnonly=[:missing_docs]`**

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,9 +1,11 @@
 [deps]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+DocumenterInterLinks = "d12716ef-a0f6-4df4-a9f1-a5a34e75c656"
 FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
 
 [sources]
-FFTW = {path = ".."}
+FFTW = { path = ".." }
 
 [compat]
 Documenter = "^1"
+DocumenterInterLinks = "^1.1"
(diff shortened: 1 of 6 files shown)


docs/make.jl
```diff
@@ -9,6 +14,8 @@ makedocs(
         "API" => "fft.md",
         "Examples" => "examples.md",
     ],
+    warnonly=[:missing_docs],
```
giordano, line 17: What's missing?

docs/Project.toml
```diff
@@ -1,5 +1,10 @@
...
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+DocumenterInterLinks = "d12716ef-a0f6-4df4-a9f1-a5a34e75c656"
+FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
+
+[sources]
+FFTW = {path = ".."}
 
 [compat]
-Documenter = "~0.26"
+Documenter = "^1"
```
giordano, line 10:
```suggestion
Documenter = "^1"
DocumenterInterLinks = "^1.1"
```

docs/make.jl
giordano: I think the docstring should be added to the documentation, since it's mentioned in the README of the package and the main page of the docs, rather than ignoring the error message.

kunzaatko: I believe as well that all the docs should be included. But this was an orthogonal issue.
For this reason, I believe it essential to build the docs with a new version of documenter, which has the capability of catching this...

giordano: Add the reference to https://github.com/JuliaMath/FFTW.jl/blob/7d0f38c58eb861d79994ec632373eea4639f8a23/docs/src/index.md?plain=1#L20

kunzaatko: Added

PR #340, opened 2026-07-10
Run FFTW in GC-safe regions

JamesWrigley:
Should help prevent GC pressure. In a larger application I observed memory not being reclaimed when doing multithreaded FFTs, which I suspect to be because we were calling FFTW without entering a GC-safe, which blocks the GC when it could be running.

Would it be possible to make a new release with this?


stevengj:
Does this seem to fix the problem in your application?

JamesWrigley:
Not sure, I wasn't able to test it yet. Claude did come up with a smaller local reproducer that did seem to be fixed though. In the real application there's lots of threads doing a lot of allocations so memory usage grew quite quickly with multithreaded plans.

PR #342, opened 2026-07-17
disallow `plan * real` for in-place complex plan

### LINKED ISSUE #341: Applying a complex in-place plan via '*' to a real vector should throw
My colleague sent me this MWE - it is about defining a complex in-place plan.
This should mutate vectors when applied via '*' as documented.
When applying this to a vector of floats it should throw, but in fact it
returns a complex vector but does not mutate the the vector.

```julia
using FFTW

plan = plan_fft!(zeros(ComplexF64, 4))
vec1 = ComplexF64[1,1,1,1]
vec2 = [1,1,1,1]

# changes vec1 as expected
plan * vec1; println(vec1,"\n")

# sends error as expected
try
    FFTW.mul!(vec2, plan, vec2)
catch e;
    println(e,"\n")
end

# does not say anything and does not change vec2
plan * vec2; println(vec2,"\n")
```

PR follows.

j-fu:
When applying a complex in-place plan to a real vector, '*' should throw similar to 'mul!'.
However, AbstractFFTs.jl defines  [`*(::Plan{T}, ::AbstractArray)`](https://github.com/JuliaMath/AbstractFFTs.jl/blob/31fc5f4de1d5a4f0f1565e879ff519d46255b02a/src/definitions.jl#L224) which copies the real vector to a `Vector{T}`  which then is passed to the  [`*(cFFTWPlan{T,K,true}, StridedArray{T})`](https://github.com/JuliaMath/FFTW.jl/blob/290d1611f3325249c0893c69f0a844182fa825b5/src/fft.jl#L835) method in FFTW.jl which now calculates and return the complex FFT for the complex vector without ever knowing that it should mutate the vector it has been passed.

Issue #341 has an MWE which shows what happens.

This PR changes the signature of `*(cFFTWPlan{T,K,true}, StridedArray{T})` to                  `*(cFFTWPlan{T,K,true}, AbstractArray)`. Now this method is directly invoked and throws as expected.  A test for this behavior has been added.

Fixes #341.

### DIFF
--- a/src/fft.jl
+++ b/src/fft.jl
@@ -833,10 +833,6 @@ for (f,direction) in ((:fft,FORWARD), (:bfft,BACKWARD))
     end
 end
 
-function mul!(y::AbstractArray, p::cFFTWPlan{T}, x::AbstractArray) where T
-    throw(ArgumentError("mul!(x,::cFFTWPlan{$(T)},y) expects x and y be of type StridedArray{$(T)}"))
-end
-
 function mul!(y::StridedArray{T}, p::cFFTWPlan{T}, x::StridedArray{T}) where T
     assert_applicable(p, x, y)
     unsafe_execute!(p, x, y)
@@ -1055,10 +1051,6 @@ function mul!(y::StridedArray{T}, p::r2rFFTWPlan{T}, x::StridedArray{T}) where T
     assert_applicable(p, x, y)
     unsafe_execute!(p, x, y)
     return y
-end
-
-function mul!(y::AbstractArray, p::r2rFFTWPlan{T}, x::AbstractArray) where T
-    throw(ArgumentError("mul!(x,::r2rFFTWPlan{$(T)},y) expects x and y be of type StridedArray{$(T)}"))
 end
 
 function *(p::r2rFFTWPlan{T,K,false}, x::StridedArray{T,N}) where {T,K,N}
--- a/src/fft.jl
+++ b/src/fft.jl
@@ -829,6 +829,10 @@ for (f,direction) in ((:fft,FORWARD), (:bfft,BACKWARD))
     end
 end
 
+function mul!(y::AbstractArray, p::cFFTWPlan{T}, x::AbstractArray) where T
+    throw(ArgumentError("mul!(x,::cFFTWPlan{$(T)},y) expects x and y be of type StridedArray{$(T)}"))
+end
+
 function mul!(y::StridedArray{T}, p::cFFTWPlan{T}, x::StridedArray{T}) where T
     assert_applicable(p, x, y)
     unsafe_execute!(p, x, y)
@@ -1047,6 +1051,10 @@ function mul!(y::StridedArray{T}, p::r2rFFTWPlan{T}, x::StridedArray{T}) where T
     assert_applicable(p, x, y)
     unsafe_execute!(p, x, y)
     return y
+end
+
+function mul!(y::AbstractArray, p::r2rFFTWPlan{T}, x::AbstractArray) where T
+    throw(ArgumentError("mul!(x,::r2rFFTWPlan{$(T)},y) expects x and y be of type StridedArray{$(T)}"))
 end
 
 function *(p::r2rFFTWPlan{T,K,false}, x::StridedArray{T,N}) where {T,K,N}
--- a/src/fft.jl
+++ b/src/fft.jl
@@ -832,7 +832,7 @@ function *(p::cFFTWPlan{T,K,false}, x::StridedArray{T,N}) where {T,K,N}
     return y
 end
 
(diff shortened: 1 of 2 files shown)


stevengj:
s/inline/inplace/

test/runtests.jl
```diff
@@ -586,3 +586,48 @@ end
         AbstractFFTs.TestUtils.test_real_ffts(Array; copy_input=true)
     end
 end
+
+
+@testset "Ensure inline version throws with wrong type" begin
```
stevengj:
```suggestion
@testset "Ensure in-place version throws with wrong type" begin
```

test/runtests.jl
```diff
@@ -586,3 +586,48 @@ end
...
+    @test catched_typerror
+    
+    catched_typerror=false
+    try
+        realplan * complexvec 
+    catch e;
+        println(e,"\n")
+        catched_typerror = true
+    end
+    @test catched_typerror
```
giordano:
```suggestion
    @test_throws MethodError FFTW.mul!(realvec, complexplan, realvec)
    @test_throws MethodError complexplan * realvec

    complexvec=ones(ComplexF64,4)
    realplan = FFTW.plan_r2r!(zeros(4), FFTW.REDFT10)
    @test_throws MethodError FFTW.mul!(complexvec, realplan, complexvec)
    @test_throws MethodError realplan * complexvec 
```

test/runtests.jl
```diff
@@ -586,3 +586,18 @@ end
...
+    realvec = [1,1,1,1]
+
+    @test_throws MethodError FFTW.mul!(realvec, complexplan, realvec)
+    @test_throws MethodError complexplan * realvec
+
+    complexvec=ones(ComplexF64,4)
+    realplan = FFTW.plan_r2r!(zeros(4), FFTW.REDFT10)
+    @test_throws MethodError FFTW.mul!(complexvec, realplan, complexvec)
+    @test_throws MethodError realplan * complexvec 
+
```
stevengj, line 602: Maybe this should be ArgumentError?

j-fu:
`mul!` is a different story. It always used to throw MethodError in this case, as there is no  "wrapper" in AbstractFFTs .

src/fft.jl
```diff
@@ -503,6 +503,16 @@ function assert_applicable(p::FFTWPlan{T,K,inplace}, X::StridedArray{T}, Y::Stri
     end
 end
 
+
+function assert_applicable(p::FFTWPlan{T}, X::AbstractArray) where T
```
stevengj: shouldn't this be restricted to in-place plans?

j-fu:
Well, that case was handeled by [`*(::Plan{T}, ::AbstractArray)`](https://github.com/JuliaMath/AbstractFFTs.jl/blob/31fc5f4de1d5a4f0f1565e879ff519d46255b02a/src/definitions.jl#L224) where a copy is created.

But I agree that testing for inline in `assert_applicable` is more readable.

Ah and method dispatch based on the type parameter via defining `assert_applicable(::Plan{T,K,true}...)`  results in method ambiguities.

src/fft.jl
```diff
@@ -504,12 +504,16 @@ function assert_applicable(p::FFTWPlan{T,K,inplace}, X::StridedArray{T}, Y::Stri
 end
 
 
-function assert_applicable(p::FFTWPlan{T}, X::AbstractArray) where T
-    throw(ArgumentError("In-place plan  FFTWPlan{$(T)} cannot be applied to a $(typeof(X))"))
+function assert_applicable(p::FFTWPlan{T,K, inplace}, X::AbstractArray) where {T,K, inplace}
```
stevengj, line 507: Instead of an `if inplace` stqtement, why not just make only to `p::FFTWPlan{<:Any,<:Any,true}`?

stevengj: Hmm, we need to make sure that the parent `assert_applicable` is called when appropriate too…

src/fft.jl
```diff
@@ -819,6 +833,10 @@ for (f,direction) in ((:fft,FORWARD), (:bfft,BACKWARD))
     end
 end
 
+function mul!(y::AbstractArray, p::cFFTWPlan{T}, x::AbstractArray) where T
+    throw(ArgumentError("mul!(x,::cFFTWPlan{$(T)},y) expects x and y be of type StridedArray{$(T)}"))
+end
```
stevengj: Why is this method necessary?  If you just delete this method won't it throw a `MethodError`?

j-fu: Yes, removing results in MethodError.  It is a question of taste what to do here, at the end, perhaps less code is better.

src/fft.jl
j-fu: Well method dispatch on inline=true resulted in some method ambiguities which I couldn't resolve easily without hitting other code paths in the tests.

